PR #79, opened 2015-01-07
MUMUP-1458 : Separate frame from home

timlevett:
So we have this neat thing written in angular.  Other teams have expressed interest in utilizing a similar frame for there apps that may live in MyUW.  We needed to separate our frame from the home stuff (layout, marketplace). This change does that.

### High Level Notes
+ Separate into 2 maven modules
+ home depends on frame
+ home is /web in the end
+ frame is /frame in the end
+ frame is still dependent on the portal for the username but that would be easy to remove that restriction if one was so bold
+ I pealed apart everything layout from main and put it in its own module. 
+ While reviewing I suggest looking at the 2nd and 3rd commits for the real meat of the change (the first commit is just moving almost everything to frame)

### Known Issues
+ Search doesn't work on frame, but works on home


vertein:
:+1: I would still like search to work.

apetro:
Awesome.  Love the progress.  Would rather not accept the search regression since presumably that would mean that a `uw-master` including this changeset couldn't be promoted to prod without a fix.  Feasible to make frame search work in this changeset?

timlevett:
Ok, to be *very* clear as I think there is confusion. Search still works for /web aka home. However, if you deployed *just* the frame, then it doesn't work.  Please let me know if I'm missing something.

apetro:
Heh.  I relied on my reading of the Known Issues rather than just trying it.  :+1: for merge accepting the gap (and the reliance on username, and..)

PR #81, opened 2015-01-08
MUMUP-1472 : angularjs-portal is now versioned and deployable using release plugin

timlevett:
Releases to uw-releases and uw-snapshot on code.doit


apetro:
:+1:

PR #82, opened 2015-01-08
LUNCH : bring back notification demo in beta settings

timlevett:
Remember that one time we had downtime and we started working on notifications, but then hid it because it wasn't ready for beta 1?  Lets un-hide this work with a beta flag.

image

image

image


apetro:
:+1:

PR #83, opened 2015-01-08
Use bower for front-end dependencies, remove old files

jhanstra:
We started installing stuff with Bower awhile ago, but never actually used them on the page. Now index.html gets the JS for libraries and frameworks from the `bower_components/` folder instead of `js/` or `js/lib/` folder.

I also added ngStorage, jQuery UI, and angular-sanitize to Bower, hence why this pull is so ginormous.

Refresher on Bower: http://bower.io/


timlevett:
I noticed `ui-bootstrap` and `sortable` were not in bower_components. Any reason for that? Guessing its because they are not in Bower, but not sure.

Assuming everything is up to date as well?  We were on AngularJS 1.3.2 I believe.

:+1:

vertein:
:+1: I don't know how to review this.  I'll hit merge after you answer @timlevett question.

jhanstra:
Yep, I couldn't find `sortable` in bower and there were 3 versions of `bootstrap-ui`, 2 of which didn't have the entire minified file (they all had a zillion components), and 1 of which I couldn't get to work. So I gave up and left the files as is :/

And yep, all updated.

PR #84, opened 2015-01-09
MUMUP-1437: Prof pic to fix long name

jhanstra:
So, I don't like doing things twice. So given the choices in the JIRA for MUMUP-1437 between fixing the long name in the header and then redoing that fix again if/when we put it in the sidebar, I propose we just put it in the sidebar AND put in a small profile picture like we've talked about in the past.

Here's how it would look:
image

Until we get the back-end in place to allow actual profile pics uploaded by users (if that's something we want to do), we could just put in a placeholder image or Bucky, like so:
image

And, this fixes the name issue, even for atypically long names :)
image

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/partials/sidebar-left.html
+++ b/angularjs-portal-frame/src/main/webapp/partials/sidebar-left.html
@@ -4,7 +4,7 @@
       <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="left" popover-trigger="mouseenter" popover-popup-delay="600"></span>
     </div>
     <div class="profile-pic">
-      <img src="img/jared.jpeg">
+      <img src="img/terrace.jpg">
       <username></username>
     </div>
     <hr>
--- a/angularjs-portal-frame/src/main/webapp/partials/sidebar-left.html
+++ b/angularjs-portal-frame/src/main/webapp/partials/sidebar-left.html
@@ -4,7 +4,7 @@
       <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="left" popover-trigger="mouseenter" popover-popup-delay="600"></span>
     </div>
     <div class="profile-pic">
-      <img src="img/terrace.jpg">
+      <img src="{{$storage.profileImg}}" ng-if="$storage.sidebarShowProfile">
       <username></username>
     </div>
     <hr>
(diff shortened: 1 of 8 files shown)


angularjs-portal-frame/src/main/webapp/partials/sidebar-left.html
```diff
@@ -1,6 +1,13 @@
 <div>
   <nav class="side-bar-nav">
-    <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="left" popover-trigger="mouseenter" popover-popup-delay="600"></span>
+    <div class="hide-sidebar-div">
+      <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="left" popover-trigger="mouseenter" popover-popup-delay="600"></span>
+    </div>
+    <div class="profile-pic">
+      <img src="img/jared.jpeg">
```
apetro: Is the file `jared.jpeg` meant to be in this pull request?  I don't think it's in [the `img` directory](https://github.com/UW-Madison-DoIT/angularjs-portal/tree/MUMUP-1437/angularjs-portal-frame/src/main/webapp/img)?

jhanstra: whoooops, forgot to take that out. just added it to sample what an actual profile pic would look like.

apetro:
I like it a lot.

I'm not sure how I feel about whether this is good-to-go with the placeholder rather than with per-user profile pics.  It *feels* like it wants to be a profile pic.  If we're okay with hinting at profile pics without having them yet, great.

If not, maybe start with something more like

> Hi, __Name__!  ([That's not me.](https://my.wisc.edu/portal/Logout)).

?

That's also a pattern that's used elsewhere and users can relate to, it still gets the name into the sidebar where it can wrap better rather than in the header, and bumping the name over there still is a step towards adding a profile pic above it, just less setting image expectations we can't yet fulfill.

vertein:
Awesome.  Way better placement for the name opposed to the name bar from the 1840's [Fashion from the time long forgotten](http://upload.wikimedia.org/wikipedia/commons/8/8b/George_Francis_Robert_Harris_by_Richard_Beard,_1840s.jpg).  :+1: 

I like the placeholder for the picture at this time.  Maybe we could turn it on/off with the settings page until we get the actual source?

jhanstra:
Leaving this until we decide what we want to do. Maybe talk about it at grooming today.

timlevett:
I'm also a :+1: . It would be nice to have the profile pic as a beta setting.

apetro:
I'm a +0.  In general I like it.  In specific I wouldn't merge this as-is; I'd rather not show what feels like a profile image when it may be some time before we can deliver on that feature.  But I'm not interested in standing in the way if there's otherwise consensus to move forward with this increment of change.  Agree that worth discussing at grooming (and worth input from @jimhelwig and/or @prtreige.

PR #87, opened 2015-01-14
Sidebar dynamic

timlevett:
During MUMUP-1497 I realized that it would be nice if the sidebar was a bit more flexible. So... I did that.

image


apetro:
Neat. :+1:

PR #90, opened 2015-01-15
Set -Djava.awt.headless=true on mvn invocations

apetro:
to avoid stealing focus under MacOS.


timlevett:
I think you want this to merge into `uw-master` not `master`

timlevett:
nm, wrong project ;P

PR #91, opened 2015-01-15
Add a ridiculously bad build-every-minute script.

apetro:
Adds a ridiculously bad shell script to build every minute.

Intended as a starting point for collaborating on something that would, say, only build on changes.

The thing is, I'm already finding this ridiculous script helpful to my flow.

### DIFF
--- a/build-every-minute.sh
+++ b/build-every-minute.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+while :
+do 
+
+./build.sh
+
+echo "Waiting one minute and then building and redeploying again."
+echo "Press [CTRL+C] to stop.."
+echo ""
+sleep 60
+
+done


timlevett:
have you ever heard of `crontab`?

apetro:
Yup.  If this was all this script was ever going to be, well, then it would stay ridiculous.  Again, intended as a starting point for collaborating on something better.

jhanstra:
So this will automatically run mvn tomcat7:redeploy in both frame and home every minute with any new changes made?

apetro:
@jhanstra yes.  With any new changes made, and without any changes if there aren't any.

Which, as @timlevett points out, is almost exactly what you'd get if you scheduled the `build.sh` on a `crontab`.  (Subtly different in that this waits a minute between runs, whereas a naive `crontab` would be about scheduling build starts regardless of duration.)

jhanstra:
:+1:

vertein:
Appears at worse case, I won't run this.

PR #92, opened 2015-01-15
MUMUP-1379 Add Beta setting for using static content card template.

apetro:
Adds a Beta setting for turning on using the static content card template rather than the default card template for portlets with associated static content.

static-content-cards-beta-setting


jhanstra:
Cool :+1:

timlevett:
:+1: . Did we want to add that variable to the default for local storage? (in `main-controller.js` line 8)

apetro:
Added default for `staticContentOnHome` in `main-controller.js` to be more like the other Beta settings.  I did test that it works properly if that `localStorage` key-value pair isn't present at all, though.  :smile:

PR #93, opened 2015-01-15
Use UW-UI-Toolkit in /web, including Bootstrap

jhanstra:
Not sure we need Bootstrap in Bower, now that UW UI Toolkit includes it. But leaving there for now.


apetro:
:+1:

PR #94, opened 2015-01-15
Minor fix for remove app & sidebar

timlevett:
+ Fixed brokenness with deleting a portlet not updating the cache of marketplace entries
+ fixed 404 for an Angularish src url.


apetro:
:+1:

PR #95, opened 2015-01-16
Cleanup to make things play nice

jhanstra:
Some small changes so that /web plays nice with the UW UI toolkit and so that undesirable styles aren't injected into /portal.


apetro:
:+1:

angularjs-portal-frame/src/main/webapp/css/buckyless/general.less
```diff
@@ -109,3 +109,11 @@ button:focus {
...
     border : 1px gainsboro solid; padding: 1em;
 }
+
+[class*="col-"] {
+  p, li {
+    a {
+      font-weight: 400;
+    }
+  }
+}
```
timlevett: Do we want to namespace this?

PR #97, opened 2015-01-16
Add commented-out example of MacOS notification and sound on build.

apetro:
Adds documentation to end of `build.sh` with a command that can be un-commented to get a MacOS notification and sound upon `build.sh` completion.

Using this locally so I notice when the build finishes and am reminded to get back to testing whatever it was that built.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,3 +7,7 @@ popd
 pushd angularjs-portal-home
 mvn -Djava.awt.headless=true tomcat7:redeploy
 popd
+
+# On MacOS this will display a notification.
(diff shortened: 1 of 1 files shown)


timlevett:
Might be neat if it checks which `OS` the user is running then uses the appropriate notification system to alert.
```
if os == tux
  notify-send x
else if os == macos
  osascript
else 
  alert "sorry you have to use windows"
```

apetro:
@timlevett That would definitely be neat.  Had the thought but didn't go after it.  Maybe go after that in a follow-on changeset?

jhanstra:
:+1:

apetro:
Superseded by #112  .

PR #99, opened 2015-01-16
Fix static content card so it works again

timlevett:
Now that we have an option for static content, lets make it work again.

There was some class changes and styling changes which made it not work so well. This change cleans that up.

screen shot 2015-01-16 at 10 58 05 am


apetro:
:+1:

PR #100, opened 2015-01-16
Remove unwanted settings controller and cleanup notification bell flag

timlevett:
+ We didn't need the settings controller since we are just setting `$storage` items that is already set on the higher up MainController scope. So I removed it.
+ Fixed the notification bell flag so it doesn't take a page refresh to see the change.


apetro:
:+1:

PR #101, opened 2015-01-16
`mvn clean` in `build.sh`.

apetro:
Sad but true, need to `clean` to avoid keeping around removed stuff.


timlevett:
Hmm, not sure i wanted this. You could have just done a mvn clean, then ran the ./build.sh. because majority of the time you don't need to rebuild the less.

PR #104, opened 2015-01-20
Cleanup

jhanstra:
Fixed:
image

And some other clean up after pulling in UW UI Toolkit.

Still Broken:
Clicking on a nav item on mobile does not collapse the navbar. Guessing this is from @timlevett adding the sidebar controller, and the `navbarCollapsed` variable is in header controller. Will talk to @timlevett when he returns about the best way to fix this. I would recommend we merge this PR and fix the issue in a separate PR.


vertein:
:+1:

timlevett:
:+1:

PR #105, opened 2015-01-21
Add `to_trusted` filter.

apetro:
Adds a `to_trusted` filter which makes it (dangerously) easy to mark content as "trusted" from AngularJS's perspective.

## How to use

Here's how you'd use this in say the `pithy-card.html` template so that the HTML from the entity files can include `JavaScript`.

```
<div ng-bind-html="portlet.pithyStaticContent | to_trusted"></div>
```

Without piping through `to_trusted`, the pithy content can only be static `HTML` that doesn't include `JavaScript`.  Safety.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/js/filters.js
+++ b/angularjs-portal-frame/src/main/webapp/js/filters.js
@@ -51,4 +51,19 @@
         return filtered;
       };
     });
+    
+    /* WARNING: THIS FILTER IS DANGEROUS.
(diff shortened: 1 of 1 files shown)


apetro:
Note that the actual `pithy-card.html` template does not include the `| to_trusted`, even with this changeset.  I don't think it should, until we're more sure we want to go in this direction and are prepared to be suitably careful about it.

vertein:
:+1:

timlevett:
:+1:

PR #106, opened 2015-01-21
MUMUP-1508 : Fix sidebar collapse on click of an element

timlevett:
Weirdness with a `ng-init` variable with attempting to access it in a child controller instance. Moved the variable to a variable on the header controller, accessing it that way works fine. #scopeIsWeird


apetro:
:+1:

PR #107, opened 2015-01-21
MUMUP-1424 : Bug: MyUW title underlining

Daha5:
https://jira.doit.wisc.edu/jira/browse/MUMUP-1424

The MyUW Beta link on the top left that takes the user to the MyUW Home was underlining, so I disabled that under every condition and now the link should never underline.

Before (link underlined):

underl

After (link not underlined):

notunderl


apetro:
:+1:

timlevett:
:+1:

PR #108, opened 2015-01-22
MUMUP-1427 : add in error message for noscript

timlevett:
Before
image

After
image

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/index.html
+++ b/angularjs-portal-frame/src/main/webapp/index.html
@@ -23,7 +23,7 @@
     <div class="container">
       <i class="fa fa-2x fw fa-exclamation-triangle pull-left"></i>
       <strong>
-        Please <a href="http://enable-javascript.com" target="_blank">enable Javascript</a> to interact with all forms and features on our website.  For further assistance, contact the <a href="http://kb.doit.wisc.edu/helpdesk" target="_blank">DoIT Help Desk</a>.
+        Please <a href="http://enable-javascript.com" target="_blank">enable Javascript</a> to interact with all forms and features on our website.  For further assistance, contact the <a href="https://kb.wisc.edu/helpdesk/" target="_blank">DoIT Help Desk</a>.
       </strong>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


timlevett:
plz wait on this. turns out they wanted diff styling.

apetro:
Nitpicky item of style, but I'd strongly favor "use" over "utilize". "utilize" is "use" with the connotation of "using other than for its intended purpose", which isn't what we're going for here.  (This is a subtle "voice" item that maybe belongs in the style guide.)

angularjs-portal-frame/src/main/webapp/index.html
```diff
@@ -18,6 +18,7 @@
 </head>
 
 <body ng-controller="MainController as mainCtrl">
+<noscript>The new MyUW requires JavaScript to run. Please enable it to fully utilize this website. Please contact the help desk with any questions/concerns.</noscript>
```
apetro: Do we believe the old MyUW doesn't require `JavaScript`?  If so worth giving the view an invitation to "Try to use Classic MyUW, which may work better than new MyUW without JavaScript."?  If not, worth making this just "MyUW requires JavaScript..." so as to not raise false hopes?

timlevett:
I think you missed the whole "plz wait on this" part @apetro . Ready for review now.

angularjs-portal-frame/src/main/webapp/index.html
```diff
@@ -18,6 +18,16 @@
 </head>
 
 <body ng-controller="MainController as mainCtrl">
+<noscript>
+    <div class="alert alert-warning alert-dismissible" role="alert" style="margin-bottom:0;">
+    <div class="container">
+      <i class="fa fa-2x fw fa-exclamation-triangle pull-left"></i>
+      <strong>
+        Please <a href="http://enable-javascript.com" target="_blank">enable Javascript</a> to interact with all forms and features on our website.  For further assistance, contact the <a href="http://kb.doit.wisc.edu/helpdesk" target="_blank">DoIT Help Desk</a>.
```
apetro: `http://kb.doit.wisc.edu/helpdesk` doesn't work for me.  I get a blank screen, under both [Chrome](http://cl.ly/image/1N0P2J3J1W2P) and [Safari](http://cl.ly/image/0E1u391t2z2M).

apetro:
:+1:

vertein:
Good use of no-script.  :+1:

PR #111, opened 2015-01-22
MUMMNG-1091 : Add my profile option to sidebar

timlevett:
image

image


jhanstra:
:+1:

apetro:
This changeset doesn't set a default value for `myProfileOption` [in the `MainController`](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/2ea76821a627f85555c262260b2cff53548233e9/angularjs-portal-frame/src/main/webapp/js/main/main-controllers.js#L17) the way other beta options are defaulted.

But this will work just fine without setting a default.

And maybe we're better off not cluttering non-beta-settings-hipster users with localstorage flags that are just the default?  We might eventually regret having set a default value when we want to switch the default value for a setting.  Probably cleaner to only set the flags when they've been set via that UI, that is, when the user expressed a preference?

timlevett:
I agree I didn't add in a default. Is the suggestion to add it or are we deciding it doesn't matter?

apetro:
I'm fine with it not mattering; mostly just noting this is different than the others; maybe the others should be more like this.

PR #112, opened 2015-01-23
Add optional parameters to build and add in mac/linux notifications

timlevett:
You can now do `./build.sh` just like before that calls `mvn clean package` OR you can now pass in what build steps you want it to do, for example `./build.sh clean install` will call `mvn clean install`.

I also added in notifications for mac and linux. The idea spawned from @apetro 's pull #97.


apetro:
:+1:

timlevett:
Changed default to `clean install`

jhanstra:
:+1:

PR #113, opened 2015-01-23
Add btn-default and btn-primary

jhanstra:
Few more changes to account for new btn-default and btn-primary classes.


apetro:
:+1:

timlevett:
:+1:

PR #114, opened 2015-01-23
Add mvn install to build.sh

jhanstra:
Should we add `mvn install` to the build.sh file? I'm a maven newb but I consistently can't get styles to transfer from frame to home when developing without running `mvn install`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,6 +4,7 @@ mvn -Djava.awt.headless=true clean package
 pushd angularjs-portal-frame
 mvn -Djava.awt.headless=true tomcat7:redeploy
 popd
+mvn install
 pushd angularjs-portal-home
 mvn -Djava.awt.headless=true tomcat7:redeploy
 popd


build.sh
```diff
@@ -4,6 +4,7 @@ mvn -Djava.awt.headless=true clean package
 pushd angularjs-portal-frame
 mvn -Djava.awt.headless=true tomcat7:redeploy
 popd
+mvn install
```
apetro: since `install` implies `package`, probably cleaner to just make the first `mvn` line

```
mvn -Djava.awt.headless=true clean install
```

timlevett:
This seems do duplicate a lot of what #112 is doing.

apetro:
I think it's nuancing what the default behavior should be.  I'm thinking `clean install` is a better default behavior since it's more likely to make a change take effect as intended.

timlevett:
Sounds good, lets close this and update #112

PR #115, opened 2015-01-23
MUMUP-1441: 'Show all' doesn't display hint

jhanstra:
Works:
image


apetro:
:+1:

timlevett:
:+1: Curious if people will get annoyed that the popover is popping over the button on which they need to enable said feature.

apetro:
Good catch, @timlevett .  Yes, that's annoying.

PR #117, opened 2015-01-28
Static content max angular view

timlevett:
In chats with the team we thought it would be an interesting idea to have static content on an angular page (I think we can credit @vertein with the original idea). This is a beta-setting level view on that.

screen shot 2015-01-27 at 11 13 28 pm

Please note this takes precedence over both static content on home and pithy content.

When you click on a card when new setting is enabled you are redirected to `/web/#/static/:fname`. This then loads another view.

screen shot 2015-01-27 at 11 08 47 pm

### Bottom line
+ This feature saves a server hit and 2 xslt transformations
+ The only real difference the user will see is speed improvement
+ #incrementalImprovement

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/js/main/main-controllers.js
+++ b/angularjs-portal-frame/src/main/webapp/js/main/main-controllers.js
@@ -11,7 +11,8 @@
       homeImg : "img/square.jpg", 
       sidebarShowProfile: false, 
       profileImg: "img/terrace.jpg", 
-      notificationsDemo : false, 
+      notificationsDemo : false,
+      staticContentMax : false,
       staticContentOnHome : false,
       pithyContentOnHome : false } );
   } ]);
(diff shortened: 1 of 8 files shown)


apetro:
Rather than implementing this as an on-off-switchable Beta feature, what if it were just a latent feature (always available) and `portlet-definition`s opt in to participating in it by setting an `alternativeMaximizedUrl` of `#/static/{{::portlet.fname}}` or so?

That is, would our existing small cards work to launch the Angular-rendered static maximized card view if those portlet-definitions set their alt max URL to the Angular path?

An advantage of that approach is we could try out this feature at the same time as building out content rendered in the small view of the cards, experiment with how those launch this, etc.

apetro:
:+1: . Progress is progress.  Faster is better.

timlevett:
That is a valid approach @apetro . That would also force us to try out each static content portlet to make sure it looked good in the new /web world before making the switch. Might be healthy even :).

Lets chat after standup and make final calls.

jhanstra:
This is a huge breakthrough, well done :+1:

apetro:
I agree.  This is a huge breakthrough and an immediately valuable (faster! snappier! better experience!) step towards more breakthroughs.

PR #118, opened 2015-01-30
Make the new simple content on angular faster by removing server side hit

timlevett:
+ No more server side hit on max static content load, for the most part
+ If you do a deep link to that portlet and a) its not in the shared service or b) another portlet is in the shared service, it then goes to the layout to get it. If the layout doesn't exist yet it calls the service to populate the layout.
+ Moved layout to `$rootScope` so we can reuse it if we stay on `/web`. Note that if you add a portlet to your favs from marketplace it clears the `$rootScope.layout` due to the change.
+ If the portlet is not in your layout we redirect to /web (could be better, could goto marketplace entry (/web/#/app/details/:fname), but that page looked a little broken, buttons didn't do actions, so I didn't do that. Another thought I had was adding fname to the search, then grabbing that and doing a search on the fname)

### DIFF
--- a/angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
+++ b/angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
@@ -3,13 +3,18 @@
 (function() {
   var app = angular.module('portal.layout.controllers', []);
 
-  app.controller('LayoutController', [ '$sessionStorage', '$scope', 'layoutService', 'miscService', function($sessionStorage, $scope, layoutService, miscService) {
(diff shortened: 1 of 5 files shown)


apetro:
:+1: .  Awesome.  Faster is better.

Agree shouldn't redirect to Marketplace entry  `/web/#/app/details/:fname` while that's a little broken.  Should backlog fixing that.

Maybe the best experience if it's not in your layout, though, would be to grab the content for the portlet as a one-off, successfully render it anyway, and still not have it in your layout.  Users should be able to Launch or deep link to portlets that aren't in their layouts (but that they do have permission to render), right?  With a way to decide they like what they see enough to Favorite it so it becomes part of their layout. :smile:

vertein:
:+1:

timlevett:
I agree @apetro . That would be ideal. Maybe if I get free time on Monday I'll build the REST service.

PR #119, opened 2015-01-30
[MUMUP-1079] Add type-ahead to the search box.

jhelmer-unicon:
Add typeahead to the searchbox.   

Screens:
uw-typeahead

Note:  I had to update ui-bootstrap to 0.12.0.  There were issues with the typeahead in 0.11.0.   I've test a fair amount and haven't found any errors, but it there may be something subtle I am missing somewhere.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
@@ -1,13 +1,14 @@
 .bucky-nav-large li:first-of-type+li{
   position:absolute;width:30%;left:35%
 }
-ul.dropdown-menu li:first-of-type+li {
-  position: relative;
-  width: inherit;
-  left: 0;
-}
-search ul.dropdown-menu {
+.portlet-search-bar ul.dropdown-menu {
   min-width: 90%;
+
+  li:first-of-type+li {
+    position: relative;
+    width: inherit;
+    left: 0;
+  }
 }
 #myuw-header span {
   font-size:25px;
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
@@ -1,6 +1,14 @@
 .bucky-nav-large li:first-of-type+li{
   position:absolute;width:30%;left:35%
 }
+ul.dropdown-menu li:first-of-type+li {
+  position: relative;
+  width: inherit;
+  left: 0;
+}
+search ul.dropdown-menu {
+  min-width: 90%;
+}
 #myuw-header span {
   font-size:25px;
 }
--- a/angularjs-portal-frame/src/main/webapp/js/main/main-directives.js
+++ b/angularjs-portal-frame/src/main/webapp/js/main/main-directives.js
@@ -18,12 +18,35 @@
     }
   });
 
-  app.directive('search', function() {
+  app.directive('search', ['miscService', 'marketplaceService', '$location', function(miscService, marketplaceService, $location) {
     return {
       restrict : 'E',
-      templateUrl : 'partials/search.html'
+      templateUrl : 'partials/search.html',
+      controller: function($scope) {
+          $scope.initialFilter = '';
+          $scope.filterMatches = [];
+          $scope.portletListLoading = true;
+          marketplaceService.getPortlets().then(function(data){
+              $scope.portlets = data.portlets;
+              $scope.portletListLoading = false;
+          });
+
+          $scope.$watch('initialFilter', function(newVal, oldVal) {
+              if (!newVal || !$scope.portlets) {
+                  $scope.filterMatches = [];
+                  return;
+              }
+
+              $scope.filterMatches = miscService.filterPortletsBySearchTerm($scope.portlets, newVal);
+          });
+
+          $scope.onSelect = function(portlet) {
+              $scope.initialFilter = portlet.name;
+              $location.path("/apps/search/" + $scope.initialFilter);
+          };
+      }
     }
-  });
+  }]);
 
   app.directive('username', function() {
     return {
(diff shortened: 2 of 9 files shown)


angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
```diff
@@ -1,6 +1,14 @@
 .bucky-nav-large li:first-of-type+li{
   position:absolute;width:30%;left:35%
 }
+ul.dropdown-menu li:first-of-type+li {
```
timlevett: We would want to give a more specific class than this I would think

angularjs-portal-frame/src/main/webapp/js/main/main-directives.js
```diff
@@ -18,12 +18,35 @@
...
+  app.directive('search', ['miscService', 'marketplaceService', '$location', function(miscService, marketplaceService, $location) {
     return {
       restrict : 'E',
-      templateUrl : 'partials/search.html'
+      templateUrl : 'partials/search.html',
+      controller: function($scope) {
+          $scope.initialFilter = '';
+          $scope.filterMatches = [];
+          $scope.portletListLoading = true;
+          marketplaceService.getPortlets().then(function(data){
```
timlevett: There is opportunity to utilize the ng-storage of the marketplace here. We should first check the cache before pulling from server side. Also remember to check the session key matches. See marketplace controller for details. Might also just move that check to the service.

angularjs-portal-frame/src/main/webapp/partials/search.html
```diff
@@ -1,6 +1,7 @@
 <form ng-submit="submit()">
   <div class="input-group">
-    <input type="search" class="main-search" placeholder="Search MyUW" name="initialFilter" ng-model="initialFilter" aria-label="Search MyUW" focus-me="showSearchFocus">
+    <input type="search" class="main-search" placeholder="Search MyUW" name="initialFilter" ng-model="initialFilter" aria-label="Search MyUW" focus-me="showSearchFocus"
+            typeahead="title as match.title for match in filterMatches | limitTo:10" typeahead-on-select="onSelect($item, $model, $label)" typeahead-loading="portletListLoading" typeahead-min-length="2">
```
timlevett: Interesting thing that came up in review. It may be interesting to have a different search for autocomplete that only searched on the name/title.

timlevett:
Overall I like this feature and the implementation. A couple suggestions inline. Also, it may be worth putting a beta setting flag (idea from @apetro ) for this for now so we can try it out in our test environment. See /web/#/settings for examples how to do that.

jhelmer-unicon:
Thanks Tim.    I'll try to circle back to this early this week and see what I can do to address your comments.

jhelmer-unicon:
Updated. 

- cleaned up the LESS a bit
- cached the getLayout() call in session storage.
- cache getUser() call in the page (necessary to avoid issues with checking the session key)
- enabled ability to turn the auto-complete on/off from the settings page
- only search by title for the auto-complete

Notes:
I think I've pretty well addressed all the comments here.   One think worth pointing out:   I did have to make one change to be aware of.  In some places, the code was using:

```javascript
return $http(...).success(...).error(...);
```

I've updated a couple of those places that I was working with to use:

```javascript
return $http(...).then(succesfn, errorfn);
```

there is subtle difference between the 2 and the way that the promise chaining works.   If you use the first method, you can't use the results from one promise in another promise -- it just calls the next promise with the same set of args.   Generally, I think the "then" option is better, but it's something to be aware of in case any issues arise.   I grepped the code and believe I updated any of the places these changes affect, but just wanted to make sure people caught that change.

Grr.  Just noticed the merge conflicts.   Will in the  AM.

timlevett:
@jhelmer-unicon - yeah, we ended up upgrading bootstrap-ui in another pull and did some restructuring for static content in /web. Sorry about that.

jhelmer-unicon:
:-)  np.  My bad.  I should have expected that, just didn't leave my self enough time last night to tackle.  I think I've got everything merged and working

angularjs-portal-frame/src/main/webapp/js/main/main-service.js
```diff
@@ -3,18 +3,24 @@
 (function() {
 var app = angular.module('portal.main.service', []);
 
-app.factory('mainService', function($http, miscService) {
+app.factory('mainService', ['$http', 'miscService', '$q', function($http, miscService, $q) {
```
timlevett: Looks like you didn't end up using `$q`

angularjs-portal-frame/src/main/webapp/js/services.js
```diff
@@ -31,10 +31,60 @@ app.factory('miscService', function($http, $window, $location) {
     $window._gaq.push(['_trackEvent', category, action, label]);
   }
 
+    var portletMatchesSearchTerm = function(portlet, searchTerm, opts) {
```
timlevett: This frame code is specific to marketplace. The idea behind the frame is to be generic so other applications can use it. Please move this to somewhere in the marketplace code base.

angularjs-portal-frame/src/main/webapp/partials/search.html
```diff
@@ -1,6 +1,26 @@
 <form ng-submit="submit()">
-  <div class="input-group">
-    <input type="search" class="main-search" placeholder="Search MyUW" name="initialFilter" ng-model="initialFilter" aria-label="Search MyUW" focus-me="showSearchFocus">
+  <div class="input-group portlet-search-bar">
```
timlevett: We may just want to overwrite the frame search in the home module as this is pretty MyUW specific.

angularjs-portal-frame/src/main/webapp/js/services.js
jhelmer-unicon: I'm gonna try a quick and dirty fix to this, but most likely this will need to roll into next week.  I suspect it may be a bit tricky to decouple those pieces.

I was wondering why there was this arbitrary split between "home" and "frame".  Makes more sense now.

timlevett:
Closing in favor of #132 which has all these changes plus some. Thanks for the excellent work @jhelmer-unicon

PR #120, opened 2015-02-02
MUMUP-1537 : Make simple content real

timlevett:
+ Removes beta setting for static content (both versions)
+ Adds in check for alternative max link
+ Swaps out "get full layout, loop through and pray it finds the portlet in your layout" with "get portlet" API call. If you don't have access or gave a bogus fname it redirects to home.
+ Restructured layout.html. We now have 3 states: `NORMAL`, `STATIC`, `PITHY`

`NORMAL` = no static content, or alt max URL was utilized
`STATIC` = static content exists, no max URL
`PITHY` = pithy flag set, has pithy content (*note:* this one trumps `NORMAL` and `STATIC`)

image


timlevett:
Depends on https://github.com/UW-Madison-DoIT/uPortal/pull/314

apetro:
:+1:

PR #124, opened 2015-02-04
MUMUP-1440 : Ga filter events

timlevett:
+ Adds in logging for ga events for localhosting
+ Adds in event for filters. Note it does wait until the filter is actually filtered (currently has a 250ms delay)


vertein:
How does this handle going to the marketplace page, then typing something in.  Does this capture that initial filter event?

timlevett:
@vertein - it captures every time the real time filter happens.

timlevett:
@vertein - confirmed expected behavior

vertein:
@timlevett cool beans. :+1:

PR #125, opened 2015-02-06
MUMUP-1565 : Admins can now see keywords if they flip the switch

timlevett:
### Admin flag
image
+ In beta settings

### With Keywords
image

+ Clicking link will do the search
+ Sorted alphabetically (case insensitive)

### Without
image

Minor changes
+ Moved the ng-repeats to the li instead of the ul for all of the details sections. We don't need a new ul for every item :)
+ removed $scope.$storage (which was pointing at session storage) in marketplaceController as it wasn't being used and conflicted with mainControllers $scope.$storage which was pointing at local storage

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/js/main/main-controllers.js
+++ b/angularjs-portal-frame/src/main/webapp/js/main/main-controllers.js
@@ -8,6 +8,7 @@
     $scope.$storage = $localStorage.$default( {
       showSidebar: true, 
       sidebarQuicklinks: false, 
+      showKeywordsInMarketplace : false,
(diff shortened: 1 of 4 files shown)


vertein:
:+1:

apetro:
I think I love the use of Beta Settings to gate access this -- it's certainly not what I had in mind when I wrote that Story, but it meets the underlying business need of hiding this from non-admin users to avoid adding UI complexity we're not ready for.  So, great example of articulating Stories in terms of business need creating opportunity for potentially quicker and better development decisions than might have been made if the Story had been over-specified.

PR #127, opened 2015-02-09
MUMUP-1539 : marketplace related app goes to marketplace entry

timlevett:
+ Related portlets now go to that marketplace entry by doing a search for the title

demo


jhanstra:
kewl :+1:

PR #128, opened 2015-02-10
Refactor LESS

jhanstra:
I moved some stuff around and renamed some classes to make more sense and more clearly organize some of our LESS. 

Cards.less is gone. It didn't really make sense, and had styles that influenced stuff on the marketplace page, stuff on the home page, and stuff across all of MyUW, like the maximized portlets themselves. It was all over the place. I split up these styles respectively into 1. The existing marketplace.less file, 2. A new file home.less for styles only relevant to the home page, and 3. A new file portlet-frame.less for stuff that applies to a portlet in general. This portlet-frame.less will be moved to the Style Guide in its next release.

TL;DR this is much better and makes more sense.


angularjs-portal-frame/src/main/webapp/css/buckyless/portlet-frame.less
```diff
@@ -0,0 +1,50 @@
+/* Portlet Frame */
+
+/*  
+The portlet frame is the div that houses the main content of the page. 
+It contains a header div (.portlet-header), usually with a title, description, and background image. 
+It also contains a body div (.portlet-body), that houses the content.
+*/
+
+.portlet-frame {
```
timlevett: Do we want to wrap the whole `portlet-frame.less` file in a namespace. Say .portlet-frame?

Therefore we for sure know that it will only effect things in the .portlet-frame scope?  

#TimRandomThoughts

timlevett:
:+1:  Refactor all the things! But seriously good call. Dig it. 1 minor suggestion in line.

angularjs-portal-frame/src/main/webapp/css/buckyless/portlet-frame.less
jhanstra: That's a decent idea, but we already have a my-uw namespace (that .portlet-frame is under) and i don't think we should have too many layers of namespaces. The idea for this file is just to outline the portlet window, i.e. this basic structure:
image
(this example is coming to the toolkit very soon, pull request is up)

For this reason, I don't think this file should get a ton added to it in the future, so might not need namespacing.

jhanstra:
If you're good with above comment @timlevett, I'll go ahead and merge.

vertein:
:+1:

PR #129, opened 2015-02-10
Added alert about css wohs

timlevett:
Just persisting this to master so its consistent to what is in production


apetro:
How hard would it be to know with some confidence when browser cache will time out such that this issue is gone?  As in, is Feb 20 the right end date?  Is it possible to know that, say, this will be a non-issue for everyone by Thursday?

timlevett:
Picked that date at random. 10 days sounded good. Open to change if you request

apetro:
I'm fine with the 10 days absent more context for the decision.

PR #130, opened 2015-02-10
MUMUP-1594 : Version CSS

timlevett:
+ Versions our `CSS` so we "clear the cache" for every portal release.
+ Tested out in predev, works well.
+ bullet points...


vertein:
:+1:

PR #132, opened 2015-02-12
MUMUP-1079 : Typeahead to search beta feature

timlevett:
This has all the work from #119 (many thanks @jhelmer-unicon ) plus some additional re-factoring.

### From #119:
Add typeahead to the searchbox.   

Screens:
http://goo.gl/Y4NViM

### Additional changes:
+ Restructured search to be in the marketplace code base, and in home module, for the most part
+ Created a stub search controller for frame so adopters have somewhere to start with
+ Moved the session storage caching for marketplace to the service layer (mimic of what was done in layout)
+ Fixed some minor bugs with on submit
+ Made the fetch of the marketplace only happen if the beta flag is set to true.

image

### Known bugs
+ When you select a typeahead item, it just goes to marketplace entry. (what it should do it TBD)
+ When you select a typeahead item, the focus is not correct when you hit the marketplace (this doesn't happen all the time, but sometimes)
+ #betafeature

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/header.less
@@ -1,6 +1,15 @@
 .bucky-nav-large li:first-of-type+li{
   position:absolute;width:30%;left:35%
 }
+.portlet-search-bar ul.dropdown-menu {
+  min-width: 90%;
+
+  li:first-of-type+li {
+    position: relative;
+    width: inherit;
+    left: 0;
+  }
+}
 #myuw-header span {
   font-size:25px;
 }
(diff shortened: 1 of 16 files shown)


apetro:
I'm :+1: after a close read, but this is enough change that more review seems warranted.  Maybe @jhanstra could review today before he's out for a couple well-earned vacation days next week, and @vertein could review on Monday when he returns?  

Balancing inserting delay with using review to help everyone be comfortable with some complex promise dancing.

timlevett:
@apetro Since it is a big change I would like to get it added sooner rather than later to avoid inevitable merge conflicts that has been plaguing this issue for over 2 weeks. That is fine to have more review, but do note that I too reviewed majority of this code as part of my copy over from the Unicon work.

I also suggest local testing please.

jhanstra:
I noticed that searching for 'email' and selecting Email/Calendar doesn't work - assuming because it has a slash in the name and so the url it goes to isn't recognized. But this is a small bug. Overall I think this is a fantastic feature that end users will love. I also think the refactoring is a lot better.

I do think we should revisit the functionality of the typeahead search, and make it go straight to the maximized portlet instead of the marketplace. But that sounds like a different pull request and this one is already huge. I say :+1:.

I'll let you guys decide if it's ready to merge, but I would say merge it.

timlevett:
Thanks for the quick turn around @jhanstra and @apetro . Agreed that this typeahead feature isn't complete. Once we have all the kinks worked out we will remove the beta flag.

PR #135, opened 2015-02-13
MUMUP-1470: Bigger Screenshots

Daha5:
Previously, screenshots were really tiny and users had no ability to enlarge them or really understand what content the portlet has to offer by seeing the screenshots. I used a bunch of CSS to make them a little bigger, added a border, and also added a hover feature that makes a larger version of the image pop out so the user can get a good glimpse of the content in this portlet.

Before:
before

After (with larger images in background + hover-enlarge ability):
screenshot

After (with smaller window):
screenshot 2 p1

After (with smaller window and hover-enlarging)
screenshot 2 p2

The bigger image disappears as soon as the mouse hovers off of it. 

Also, for those wondering, having the @media only screen several different times was the very inefficient way of keeping the enlarged images from being cut off, but it works very well.

Please note that this is disabled on mobile/smaller devices.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
+++ b/angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
@@ -83,10 +83,17 @@
     </div>
     <div class="col-xs-12 col-sm-9 portlet-screenshots" ng-if="portlet.marketplaceScreenshots.length !== 0">
       <h3>Screenshots</h3>
-      <ul >
-        <li ng-repeat="screenshot in portlet.marketplaceScreenshots">
-          <img ng-src="{{ ::screenshot.url }}" alt="{{ ::screenshot.captions }}">
-          <p ng-repeat="caption in screenshot.captions">{{ ::caption }}</p>
+<<<<<<< HEAD
+      
+      <ul class="enlarge" ng-repeat="screenshot in portlet.marketplaceScreenshots">
+        <li>
+          <img ng-src="{{ ::screenshot.url }}" width ="320px" alt="{{ ::screenshot.captions }}">
+          <span>
+          	<img ng-src="{{ ::screenshot.url }}" alt="{{ ::screenshot.captions }}">
+          	<p ng-repeat="caption in screenshot.captions">{{ ::caption }}</p>
+          </span>
+          
+>>>>>>> MUMUP-1470: Bigger Screenshots
         </li>
       </ul>
     </div>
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
```diff
@@ -83,10 +83,17 @@
     </div>
     <div class="col-xs-12 col-sm-9 portlet-screenshots" ng-if="portlet.marketplaceScreenshots.length !== 0">
       <h3>Screenshots</h3>
-      <ul >
-        <li ng-repeat="screenshot in portlet.marketplaceScreenshots">
-          <img ng-src="{{ ::screenshot.url }}" alt="{{ ::screenshot.captions }}">
-          <p ng-repeat="caption in screenshot.captions">{{ ::caption }}</p>
+<<<<<<< HEAD
```
timlevett: Looks like an unmerged file. Please cleanup

angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
```diff
@@ -83,10 +83,17 @@
...
+      
+      <ul class="enlarge" ng-repeat="screenshot in portlet.marketplaceScreenshots">
+        <li>
+          <img ng-src="{{ ::screenshot.url }}" width ="320px" alt="{{ ::screenshot.captions }}">
+          <span>
+          	<img ng-src="{{ ::screenshot.url }}" alt="{{ ::screenshot.captions }}">
+          	<p ng-repeat="caption in screenshot.captions">{{ ::caption }}</p>
+          </span>
+          
+>>>>>>> MUMUP-1470: Bigger Screenshots
```
timlevett: more cleanuup

angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
```diff
@@ -83,10 +83,13 @@
     </div>
     <div class="col-xs-12 col-sm-9 portlet-screenshots" ng-if="portlet.marketplaceScreenshots.length !== 0">
       <h3>Screenshots</h3>
-      <ul >
-        <li ng-repeat="screenshot in portlet.marketplaceScreenshots">
-          <img ng-src="{{ ::screenshot.url }}" alt="{{ ::screenshot.captions }}">
-          <p ng-repeat="caption in screenshot.captions">{{ ::caption }}</p>
+      <ul class="enlarge" ng-repeat="screenshot in portlet.marketplaceScreenshots">
```
jhanstra: Probably want to put the ng-repeat on the li instead of the ul to be more semantic.

jhanstra:
One small suggestion - other than that looks good!

angularjs-portal-home/src/main/webapp/partials/marketplace-portlet.html
timlevett: True.

Daha5:
I'm not exactly sure why I changed that to begin with, but I put it back on the li. 
Thanks Jared!

jhanstra:
I think this is ready to merge. @timlevett @vertein @apetro @keirserrie any objections?

timlevett:
:+1:

keirserrie:
:+1:

PR #136, opened 2015-02-16
R&D : Tim tries unit tests for angular

timlevett:
Watched a couple presentations on proper testing in Angular. Started poking around with what @jhanstra started a while back.

+ Setup a test directory (maven style) and got tests up and running.
+ Created a basic test for the main controller
+ created a `README.md` for how to set it up locally
+ Deleted `scripts.js` as it wasn't being used


angularjs-portal-frame/src/main/webapp/js/scripts.js
```diff
@@ -1,13 +0,0 @@
-var up = up || {};
-
-$(document).ready(function() {
-    $('.app-drawer').toggle(
```
vertein: Was this not needed code?

timlevett: True statement. Did a search for `app-drawer` and came up empty.

vertein:
:+1: With latest comment.

PR #137, opened 2015-02-19
MUMUP-1609 : Config on portal module

timlevett:
+ Add `app-config.js` which is a module that is injected for constants
+ Created a `conf/app-config` directory that holds `ENV-config.js` files (one for each environment)
+ Created `conf/setup_env.sh` which will copy over all the files for a given environment.
Usage example: 
```
./angularjs-portal-home/src/main/webapp/conf/setup_env.sh predev
```
+ Created test area for home, made it work, added a test to verify the flag was getting set in the controller init. Note that I stubbed out a lot of mock services and stuff.
+ Moved app-config to resources


vertein:
:+1:

jhanstra:
:+1:

PR #139, opened 2015-02-20
MUMMNG-1145 : add static page pagehit for ga

timlevett:
+ Reorganize list of services, etc..
+ add in `miscService.pushPageview()`


jhanstra:
:+1:

PR #141, opened 2015-02-20
Fix marketplace light so it isn't sortable

timlevett:
Minor change to make it so marketplace light is not sortable


apetro:
:+1: :smile:

PR #143, opened 2015-02-23
MUMUP-1610: Refactor to make widgets on separate page

jhanstra:
Layout.html has been split into two files: home-list-view.html and home-widget-view.html. Each calls the new home-header directive, has its own respective controller and its own respective content.


angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
```diff
@@ -122,6 +122,129 @@
...
+                      //index, length, movingNodeId, previousNodeId, nextNodeId
+                      var prevNodeId = ui.item.sortable.dropindex != 0 ? $scope.layout[ui.item.sortable.dropindex - 1].nodeId : "";
+                      var nextNodeId = ui.item.sortable.dropindex != $scope.layout.length - 1 ? $scope.layout[ui.item.sortable.dropindex + 1].nodeId : "";
+                      layoutService.moveStuff(ui.item.sortable.dropindex, $scope.layout.length, node.nodeId, prevNodeId, nextNodeId);
+
+                  }
+              }
+      };
+
+      this.toggleDiv = function toggleDiv(nodeId) {
```
timlevett: I think this function can be chucked.

timlevett:
:+1: for the initial split. We could refactor some of the functions that are going to be the same for both controllers to a shared service, but that can be done later.

angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
jhanstra: Chucked

jhanstra:
Merging

PR #144, opened 2015-02-23
MUMUP-1615 : Create servlet container, create configurable rest containers

timlevett:
+ Creates servlet for `/api`
+ Created a REST type end point with {key}.json where key is a preconfigured endpoint service
+ Uses basic auth or no authentication
+ Added an example properties file, must rename to have it work (will be in overlay)

+ Changed new stuff to point at servlet for an example of how we can utilize this new service.

### DIFF
--- a/angularjs-portal-home/src/main/resources/logback.xml
+++ b/angularjs-portal-home/src/main/resources/logback.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2012, Board of Regents of the University of
+    Wisconsin System. See the NOTICE file distributed with
+    this work for additional information regarding copyright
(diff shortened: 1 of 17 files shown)


timlevett:
Note this will break the build as we will need an overlay with the `endpoint.properties`

angularjs-portal-home/src/main/resources/logback.xml
```diff
@@ -0,0 +1,70 @@
...
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="LOG"/>
+    </root>
+
+    <logger name="edu.wisc" additivity="false" level="DEBUG">
+        <appender-ref ref="LOG"/>
+    </logger>
+</configuration>
```
vertein: end of file

vertein:
:+1: Let's do this

PR #145, opened 2015-02-24
Replace absolute url to crest image with relative path

nblair:
The previous iteration had an absolute (MUM deployment specific) URL for the crest image. Changed to a relative URL available in the build result.

Allows downstream consumers of angularjs-portal-frame to use this
partial unmodified.


timlevett:
:+1: :+1:

PR #146, opened 2015-02-24
MUMUP-1571: Widgets are a thing

jhanstra:
This pull implements the outline for the widget view. It is very similar to the minimized view, except that it has larger boxes ([so much room for activities!](https://www.youtube.com/watch?v=29jij4ldkNs)) and if there is pithy or widget content, it will display it. This is still a work in progress and rough around the edges, but it is a good start to getting widgets rolling.

### Notable code changes:
- Removed `<pithy-content-card>` and `<static-content-card-max>` directives, as they were sort of confusing and very similar to the `<default-card>` directive. Now there is just one directive for minimized apps: `<default-card>` and one directive for widget apps: `<widget-card>`
- In `<widget-card>`, if there is pithy content (we really need a new name for that) or even cooler widget content, it will display it. Otherwise, it will default to showing a large icon and the title, and the whole box will be clickable to launch the portlet.

### Screenshots:
The My Courses widget below is just a very simple example of how a widget would work. It's not pulling in any dynamic content yet or styled at all.
image

image


angularjs-portal-home/src/main/webapp/partials/home-list-view.html
```diff
@@ -28,18 +28,8 @@
...
-         <default-card 
-          ng-if="'NORMAL' === layoutCtrl.portletType(portlet)">
-         </default-card>
-         <static-content-card-max 
-          ng-if="'SIMPLE' === layoutCtrl.portletType(portlet)">
-         </static-content-card-max>
-         <pithy-content-card 
-          ng-if="'PITHY' === layoutCtrl.portletType(portlet)">
-         </pithy-content-card>
+         <default-card></default-card>
```
timlevett: wait, what? are we getting ride of simple?

jhanstra: nope, see '<default-card>` and '<widget-card>`

timlevett: nm, i get it.

angularjs-portal-home/src/main/webapp/partials/home-widget-view.html
```diff
@@ -25,9 +25,41 @@
     <div data-loading class='loading-gif'><img src="img/ajax-loader.gif" alt="Loading content, please wait"></div>
     
     <!-- Widget View -->
-    <div style="text-align:center; margin-top:30px">
-      <h3>Coming Soon</h3>
-      <p>Widgets are coming soon! Woohoo!</p>
-    </div>
+    <ul ui-sortable="sortableOptions" ng-model="layout" class="tile-list">
+      <!-- Widget Example -->
+      <li class="col-xs-12 col-sm-6 col-md-4 col-lg-3 no-padding list-container">
```
timlevett: If you were feeling ambisous you could have a demo ribbon like [here](https://github.com/UW-Madison-DoIT/my-profile/blob/master/my-profile-webapp/src/main/webapp/partials/basic-info.html#L2)

timlevett:
:+1:

angularjs-portal-home/src/main/webapp/partials/home-widget-view.html
apetro: Demo ribbon is a beautiful idea, helps set the right expectations about this not-yet-really-the-minimum-viable-product.  Love it!

jhanstra:
Merging

PR #147, opened 2015-02-25
MUMUP-1704 : Support "App" model

nblair:
This change set includes 2 important changes:

1. The 'app' variable provided by angular-page.js is no longer hidden
from global scope. This will allow "Apps" that overlay
angularjs-portal-frame to have access to the result of angular.module.
2. Adds 'my-app.js' to frame.html. The file included in this project is
intentionally blank. It's inclusion allows an "App" developer an
extension point; the end result is that the downstream app does not have
to have a copy of frame.html, they can just write their own my-app.js.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/frame.html
+++ b/angularjs-portal-frame/src/main/webapp/frame.html
@@ -71,5 +71,6 @@
 <script type="text/javascript" src="js/filters.js"></script>
 <script type="text/javascript" src="js/ui-bootstrap.js"></script>
 <script type="text/javascript" src="js/ga.js"></script>
+<script type="text/javascript" src="my-app.js"></script>
 </body>
 </html>
--- a/angularjs-portal-frame/src/main/webapp/js/angular-page.js
+++ b/angularjs-portal-frame/src/main/webapp/js/angular-page.js
@@ -1,4 +1,3 @@
-(function() {
  var app = angular.module('portal', [
     'ngRoute',
     'ngStorage',
@@ -21,7 +20,4 @@
       otherwise({templateUrl: 'partials/main.html'});
       }
  	]);
-
-
-
-})();
+ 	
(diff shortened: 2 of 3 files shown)


nblair:
I plan on pushing some changes to my-app-seed shortly that cover what the App developer needs to do and briefly how it works.

jhanstra:
We should probably consider changing the name of angular-page.js. Maybe to app-frame.js or app.js. Thoughts?

nblair:
README updated on app seed project: https://github.com/nblair/my-app-seed

nblair:
@jhanstra From what I've been exposed to with Angular, 'app.js' seems to be the convention.

timlevett:
app.js would be fine with me. What about the other files that are redeclaring the `var app` for each module? Should we restructure that? Or change the var app in the global scope to global-app or something?

angularjs-portal-frame/src/main/webapp/frame.html
```diff
@@ -71,5 +71,6 @@
 <script type="text/javascript" src="js/filters.js"></script>
 <script type="text/javascript" src="js/ui-bootstrap.js"></script>
 <script type="text/javascript" src="js/ga.js"></script>
+<script type="text/javascript" src="my-app.js"></script>
```
timlevett: any reason this isn't in the js directory?

nblair:
I did a quick search for other files that declare 'var app', and it looks as each of the other cases do so in an anonymous function. I'm with you partially, it feels a little odd to have an app served out of one context (web) provide 'app' only to have a completely different app use it. Seems like it should have some sort of warning label.

angularjs-portal-frame/src/main/webapp/frame.html
nblair: I could ask the same question why the other scripts are in a js directory :)

The angular style guides recommend organizing source by function, rather than by language. 'my-app.js' is seemingly the linchpin, so it's at the root of the web app. App developers would then have sub-directories by app function containing controllers/services/partials/etc.

timlevett:
:+1:

PR #148, opened 2015-02-25
MUMUP-1628, MUMUP-1638: Mockups for Email, Calendar, My Courses

jhanstra:
This is how a real widget could look. Now all that's left is to swap out some real data :)

image

image

image


timlevett:
:+1: , we may want to pull these demos out into a separate file just so they are not clouding the real content code. I know its temporary so more just random thought.

apetro:
That's a fine idea you have there, @timlevett .  Much to be said for keeping `uw-master` continuously shippable. :smile: :shipit: .

jhanstra:
Demos now pulled into separate directive and behind a beta flag

timlevett:
:+1: Excellent way to demo the new designs and good footwork for future templating

PR #150, opened 2015-02-26
MUMUP-1615 : Mail widget for realz

timlevett:
### Please Read
I know, TL;DR. I would like to have 2x :+1: for this at least. Maybe a post-standup chat.

#### Screenshots

Single Account
image

Multiple Accounts
screen shot 2015-02-25 at 10 54 46 pm

Side by Side
image

#### Features
+ Creates a `option-link` widget (idea behind it is one that has a select, and that changes the url for the link)
+ Creates a way of calling the widget URL when the widget mode is loaded (got it working, could be a little better)
+ Plugs in data when we get it
+ Adds in a `WIDGET` flag for generic widgets that will display content from callback (this could be better, but it's a start)
+ Adds in a `OPTION_LINK` flag that will use the option link partial (email will be one of those for now)
+ On selection of the option it will update the URL

#### TODO || Things that could be better
+ The promise to get the widget content is doing data.result, another email JSON specific thing
+ I'm not a huge fan of passing the index to the promise. if that takes a while a reorder could mess things up.
+ Make option link widget more generic

#### Notes
+ This depends on a uPortal PR https://github.com/UW-Madison-DoIT/uPortal/pull/320
+ This is depending on a change to the JSON we are getting from wiscmail to add in the email addresses as a field, for demo purposes we can just use the `/samples/wiscmail-accounts.json`

### DIFF
--- a/angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
+++ b/angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js
@@ -118,12 +118,9 @@
                                               
     miscService.pushPageview();
     $scope.toggle = APP_FLAGS.enableToggle;
-    if(typeof $rootScope.layout === 'undefined' || $rootScope.layout == null) {
-      
-      $rootScope.layout = [];
+    var that = this;
     
-      layoutService.getLayout().then(function(data){
-        $rootScope.layout = data.layout;
+    that.populateWidgetContent = function() {
         for(var i=0; i < $rootScope.layout.length; i++) {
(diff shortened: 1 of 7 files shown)


vertein:
I don't mind the index, we talked about the pass by reference face to face.

I do mind the specific email feed.

Other than that, this looks good.  Thanks for posting this to start the convo.

timlevett:
Updated to have a more generalized option-link directive, and added a coat of wax.

jhanstra:
Looks good! :+1:

vertein:
:+1:

PR #153, opened 2015-03-02
MUMUP-1713 : Fix marketplace-light so its not draggable

timlevett:
### Before
http://goo.gl/mVXrQq

### After
http://goo.gl/w1olhK


keirserrie:
:+1: eventually will be good to have different visual treatment for this so it doesn't look like other widgets

timlevett:
I agree, it could use some style differences. This was more of a functional fix so it doesn't break your in memory layout.

PR #160, opened 2015-03-06
Fix styling on new-stuff bar

jhanstra:
image

image

This was the easiest fix. If you have other ideas for how to display the new stuff are, let me know.

Also, since us developers are going to see this forever and some of us are OCD, I fixed the New New annoyingness :)


apetro:
Nice.  Better is better.  This is the kind of pointed fix that's right smack dab firmly in the spirit of a cleanup week.  Same thing we got, but less broken. :heart:

timlevett:
:+1:

keirserrie:
:+1:

PR #161, opened 2015-03-06
MUMUP-1729 : Bug : Stays scrolled down from bottom of favorites to somewhere in /apps

Daha5:
This was a bug where when the user would be scrolled down on their home page and click on the add to favorites (tile), then they would be taken to the marketplace but the scroll would not reset, and thus they'd be left in the middle of the page. 

Here are some screenshots from the JIRA story to help you understand the problem.

User scrolls down and decides to click on Add to favorites

before clicking

Immediately after clicking the user sees this, instead of being taken to the top of the marketplace.

after clicking

Anyways, I added some javascript that scrolls up whenever the button is clicked

Now it looks like this:

Before clicking:

fixed before clicking

After clicking:

fixed after clicking

### DIFF
--- a/angularjs-portal-home/src/main/webapp/home.html
+++ b/angularjs-portal-home/src/main/webapp/home.html
@@ -45,7 +45,7 @@
         <span class="fa fa-bars"></span>
       </div>
       <div id="region-main" class="col-xs-12 my-uw" ng-class="{'col-sm-10 col-sm-offset-2' : $storage.showSidebar, 'col-sm-11 max-view' : !($storage.showSidebar)}">
-        <div ng-view></div>
+        <div ng-view="" autoscroll=""></div>
       </div>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


timlevett:
Based on some google foo I found this: http://stackoverflow.com/questions/22779753/how-to-reset-scroll-position-on-a-route-change

Which says, just add autoscroll to the ng-view.

basically do this:
```diff
index a84dec9..b9a8075 100644
--- a/angularjs-portal-home/src/main/webapp/home.html
+++ b/angularjs-portal-home/src/main/webapp/home.html
@@ -45,7 +45,7 @@
         <span class="fa fa-bars"></span>
       </div>
       <div id="region-main" class="col-xs-12 my-uw" ng-class="{'col-sm-10 col-sm-offset-2' : $storage.showSidebar, 'col-sm-11 max-view' : !($storage.showSidebar)}">
-        <div ng-view></div>
+        <div ng-view="" autoscroll=""></div>
       </div>
     </div>
   </div>

```

timlevett:
:+1:

PR #162, opened 2015-03-10
MUMUP-1732 : Widget static content (pithy) is no longer made out of wood

timlevett:
Widget static content is no longer hidden behind a beta flag


apetro:
Stupid question: this doesn't clean up the `LocalStorage` flag where it has been locally stored, though it does make that flag have no effect.  That's good enough, right?  There's no reason to heroically clean up the stray flag hanging out in a very few `LocalStorage`s?

Or perhaps more to the point, is there anything to learn from this along those lines that will inform our future use of Beta flags and their retirement?

timlevett:
Yeah, we are not resetting people's local storage. Should be fine since we are not using said flag anywhere.

PR #163, opened 2015-03-10
MUMUP-1746 : Sticky layout selection

timlevett:
+ Creates toggle directive that manages which one you should be on
+ Adds sanity checking to make sure you are on the correct view
+ Consolidates exactly same code to home header
+ Moves page view to toggle to remove duplicate page hits
+ Resets to default if someone tries playing with localStorage
+ unit tests
+ Add in unit test setup.sh script for easy test setup


vertein:
:+1:

PR #166, opened 2015-03-11
Loading icon upgrade

timlevett:
+ Changed loading icon from an image to a `fa-icon` with `fa-spin`
+ Icon choice was voted upon: https://plus.google.com/116613726767369711410/posts/GKXALDogftd
+ Added a forever loading icon in beta-settings page so we can see it load foorrrreeeeverrrr

image

http://goo.gl/XLaS8M


keirserrie:
:+1:

PR #167, opened 2015-03-12
MUMUP-1749: Functional, not ugly marketplace details page

jhanstra:
- Rate works
- Launch works
- Add to Home works
- Link back to App Marketplace

image

image

image

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
@@ -241,14 +241,17 @@
   }
   
   ul.enlarge span{
-  	position:absolute;
-  	left: -9999px;
-  	padding: 5px 5px 0px 5px;
-  	background:#D8D8D8;
+  	position:fixed;
+    left:9000px;
+  	padding: 10px 10px 5px 10px;
+  	background:#333;
   	font-size:.9em;
   	text-align: center;
-  	color: black;
-  	-webkit-border-radius: 8px;
+  	color: #ccc;
+  	border-radius: 4px;
+    p {
+      margin:5px;
+    }
   }
 
   ul.enlarge li:hover{
@@ -258,39 +261,13 @@
   
   @media only screen and (min-width: 768px){
   	ul.enlarge li:hover span{
-  		
-  		top:-220px;
-  		left:-190px;
-  		width:590px;
+      width:60vw;
+  		top:20%;
+      left:28.67%;
   	}
   }
   
-  @media only screen and (min-width: 956px){
-    ul.enlarge li:hover span{
-  		
-  		top: -300px;
-  		left: -210px;
-  		width: 740px;
-  	}
-  }
-
- 
-  @media only screen and (min-width: 1330px){
-    ul.enlarge li:hover span{
-    	
-  		top: -220px;
-  		left:-250px;
-  		width:640px;
-  	}
-  } 
-   @media only screen and (min-width: 1557px){
-    ul.enlarge li:hover span{
-    	
-  		top: -300px;
-  		left: -210px;
-  		width: 740px;
-  	}
-  }
+  
   
 
   
@@ -387,18 +364,22 @@ input:focus,span:focus {
 .portlet-details-page {
   padding-bottom:0px;
   h2 {
-    margin:0px;
+    margin:0px 0px 10px;
     padding-top:10px;
   }
   .details-header {
     padding:10px 5px;
     border-bottom: 1px solid @color1;
-  }
-  .app-description {
-    padding:10px 20px;
+    .back-to-home {
+      color:#999;
+      &:hover {
+        color:#666;
+        text-decoration: none;
+      }
+    }
   }
   .desc-item.rating-div {
-    padding-top:30px;
+    padding-top:10px;
   }
   .rating {
     margin-left:0px;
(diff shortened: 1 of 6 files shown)


timlevett:
Looks great :+1: 

Additional Thoughts:
- Do we want a link back to apps on the top somewhere?
- Did we want to change marketplace (just /apps) to point to this details page some how? Maybe when they click the details link (which could then simplify that page a bit)

jhanstra:
@timlevett i think having it at the bottom is fine. there's not a lot of content on the page, so it should almost always be visible on desktop. maybe we could put it at the top on mobile? thoughts @keirserrie?

I like the idea to make the details link go here, along with cleaning up that page. Only downside is that you're redirected to a new page instead of having everything inline, but it's all angular so load time isn't an issue. :)

Also add to faves works now.

jhanstra:
Here's how it could look with it at the top. I kinda like it
image

timlevett:
Interesting. I like it, but I also liked the button on bottom. Both?

jhanstra:
Updated with keywords only if there are keywords, screenshots only if there are screenshots, style tweaks to @Daha5's enlarged screenshots, titles for description and ratings sections, padding between title and icon, and 'Back to Apps' in two places. 

image

image

jhanstra:
Said style tweaks to enlarged screenshots:
image

keirserrie:
Good stuff, can you show mobile, wanna make sure touch area is good for "back to apps"

Also let's use consistent left aligned treatment for back to apps on bottom

timlevett:
Looks awesome @jhanstra ! The details is really slick. 

I did notice one bug on normal `/apps` when locally testing on tablet/mobile sizes.

image

The desktop launch buttons are not hiding properly.

timlevett:
Also, you may want to rebase as i know this page has changed a bit. Looks a bit out of date.

Ratings moved over, icon updated, etc... 

You may want that for copying over to details.

angularjs-portal-home/src/main/webapp/partials/marketplace-details.html
```diff
@@ -1,3 +1,106 @@
...
+				<h3>Ratings</h3>
+				<span><rating ng-model="portlet.rating" readonly="true" class="rating"></rating>
+	    ( {{::portlet.userRated}} rating<span ng-if="portlet.userRated !== 1">s</span> )
+	  		</span><br>
+	  		<button ng-click="openRating('sm',portlet.fname, portlet.title)" class="btn btn-default rate-button">Rate</button>
+			</div>
+			<div class="col-xs-12 col-sm-4 col-lg-3 desc-item center">
+				<h3>Related Apps</h3>
+	      <ul>
+	        <li ng-repeat="related in portlet.relatedPortlets"><a href="#/apps/search/{{ ::related.title }}">{{ ::related.title }}</a></li>
```
apetro: Once the app-specific URL works, should the related app titles link to the entry for that related app rather than to search for the app by name?  (I think very much yes -- enabling the related apps link to behave in this way was a big motivator for this story for me. :smile: )  But I readily defer to Design on this detail, especially if I get a good shot at articulating why I think this should work the way I think it should. :smile: .

timlevett: :+1:

jhanstra:
@keirserrie here's mobile:
image

jhanstra:
@timlevett nice bug find. fixed and rebased.

timlevett:
The category is linkable but doesn't do anything, should that go back to search with that category selected?

jhanstra:
@apetro, I implemented new Related Links and also made 'Details' link on Marketplace go to this new details page. 

Gif:
screen recording 2015-03-12 at 04 23 pm

timlevett:
That is neat.

jhanstra:
@timlevett Yes definitely. I think I might want to make this the stopping point though and tackle that as part of streamlining marketplace. The position of that link/button could change.

keirserrie:
i think that back to apps link will work for fat fingers :+1:

angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
```diff
@@ -383,4 +360,70 @@ input:focus,span:focus {
 .rating-input span {
     font-size: 1.5em;
     cursor: pointer;
+}
+.portlet-details-page {
```
timlevett: :+1: for namespacing

angularjs-portal-home/src/main/webapp/partials/marketplace-details.html
```diff
@@ -1,3 +1,106 @@
...
-  <marketplace-portlet></marketplace-portlet>
-</div>
+<div ng-controller="MarketplaceController as marketplaceCtrl">
+	<div ng-controller="MarketplaceDetailsController as detailsCtrl" class="portlet-frame col-xs-12 col-md-12 portlet-details-page">
+	  
+	  <div class="row details-header no-margin">
+	  	<div class="col-xs-12 col-lg-6">
+	  		<h2>
+			    <span class="fa fa-circle" ng-show="portlet.faIcon === null"></span>
+			    <span class="fa {{::portlet.faIcon}}" ng-show="portlet.faIcon != null"></span>
```
timlevett: Could we utilize the portlet-icon here?

timlevett:
Noticed this code in the marketplace-controller.js
```js
if($routeParams.fname !== null) {
  $scope.showDetails = true;
};
```

Prob don't need that anymore.

Also, we are calling marketplaceServices.getPortlets() twice :sake: .  I think we just need the first one. Thoughts @jhanstra ?

jhanstra:
Great suggestions @timlevett. Updated addressing both comments.

timlevett:
:+1: :shipit:

PR #170, opened 2015-03-13
MUMPO-433: Screenshots don't resize if user is not in chrome

Daha5:
This pull request fixes the problem with screenshots not re-sizing in Firefox. I set the width for the default screenshots to 100% and whenever the window is >= 768 pixels the width of the default screenshots automatically changes to 320px (so they don't get too big for desktop windows). Also I set the width of the enlarged screenshots to 100% so that it doesn't get affected by the width change to 320px.

Before and After:                                                                                                                                               

before screenshots small screenshots                                 

Also, because users cannot get an enlarged version of the screenshots when their devices width is below 768px, I had the default screenshots width set to 100% only under that scenario (width<768px), so that tablet /mobile users can enjoy the bigger screenshots and so that the screenshots take up the otherwise empty space.

See below (Note that this does not maximize on hover):

tablet size

All screenshots were taken in Firefox.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
@@ -236,6 +236,7 @@
   
   ul.enlarge span img{
   	padding: 2px;
+  	width:100%;
   }
   
   ul.enlarge span{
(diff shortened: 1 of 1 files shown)


angularjs-portal-frame/src/main/webapp/css/buckyless/marketplace.less
```diff
@@ -256,13 +257,19 @@
   	cursor:pointer;
   }
   
+  img{
+  		width:100%;
+  }
```
timlevett: Please consolidate this with the other selector for img in this file that is under the screenshots
```
img {
    max-width:100%;
  }
```

timlevett:
@Daha5 - please fix the merge conflicts so we can merge. :+1:

Daha5:
Merge conflicts should be good now...

timlevett:
:+1:

PR #171, opened 2015-03-16
MUMMNG-1262 : Add in isEmpty flag to genericController

timlevett:
+ Adds in isEmpty flag so generic template widgets can utilize the loading-gif directive
+ Adds in config option `evalString` which would be a string that would be evaluated during a return call for emptiness.


vertein:
spin forever :shipit:

jhanstra:
:+1:

timlevett:
Related pull: https://github.com/UW-Madison-DoIT/uPortal/pull/331

PR #172, opened 2015-03-17
with FA UPDATE FA version to: 4.3<=version<5

vertein:
Was shown a little bower foo by @jhanstra.  Version was set to use 4.2.X release.  Changes to use 4.X.X release.  FA uses semantic versioning, guaranteeing (as much as one can) that breaking backwards compatibility requires a major release.  Minor releases are reserved for new icons and new additions.


timlevett:
:+1:

PR #173, opened 2015-03-17
Styles for Widgets

jhanstra:
Some reusable styles for widgets, coming soon to a style guide near you. And an example for how to layout a widget.


timlevett:
:+1:

apetro:
I :heart: the idea of getting to re-usable styles and style guidance.  I anticipate we're going to continue to need to add and update widgets, and driving down the marginal cost of doing that would be lovely.  `#embracechange`.

PR #176, opened 2015-03-19
MUMUP-1717 : Move help to /web

timlevett:
Help is static content, changing to use the #/static/:fname


vertein:
:+1: for the change :+1: for the new end of line

PR #177, opened 2015-03-20
MUMUP-1770 : auto complete selection now goes to details

timlevett:
+ If one finds a result in the autocomplete that they like and they select it, it goes to the details page
+ If one just hits enter or the search button on the search, then it goes to marketplace with that filter

#### gif of goto details
http://goo.gl/i9p3QR

#### gif of goto search results
http://goo.gl/EqITxe


keirserrie:
:+1:

apetro:
Nicely done!  I hadn't thought about when one would expect this to still be search, but now that I see this, I think this got it exactly right.

jhanstra:
Late to the game but this is awesome :+1:

PR #178, opened 2015-03-24
Adds uw-ui-toolkit to bower.json

vertein:
Adds the toolikit to bower.json


jhanstra:
:+1: for uw-ui-toolikit

PR #181, opened 2015-03-25
Update README to prevent future issues

jhanstra:
Does this make sense here? Will this minimize issues in the future? The lack of a manager role seems to be the problem I was having.


timlevett:
:+1:

apetro:
I do have the `manager` role listed in mine, and mine works; news to me that it was required, but at the least not gonna hurt anything; :+1: .

PR #185, opened 2015-04-06
Use RequireJS to make my-app more extensible

paulerickson:
With all the scripts included in frame.html, it is easy to overlay the single my-app.js file, but hard to accommodate a modular app with many js files.

With this change, a downstream application can use require to pull in their own dependencies within their overlaid my-app.js (which is now a require module).  Angular will initialize only afterward (this is in lieu of the ng-app directive).

If this works well, a good next step would be to isolate the rest of the javascript resources behind require, so that the frame and app can be organized into their own modules with no risk of interfering with each other through the global namespace.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/main.js
+++ b/angularjs-portal-frame/src/main/webapp/main.js
@@ -1,0 +1,59 @@
+require.config({
+
+  paths: {
+    'my-app': 'my-app.js'
+    /* TODO: Incorporate the <scripts/> that are being globally imported in frame.html
+    angular: "bower_components/angular/angular.min.js",
+    config          : "js/config.js",
+    jquery          : "bower_components/jquery/dist/jquery.min.js",
+    jqueryUi        : "bower_components/jquery-ui/jquery-ui.min.js",
+    angularRoute    : "bower_components/angular-route/angular-route.min.js",
+    angularSanitize : "bower_components/angular-sanitize/angular-sanitize.min.js",
+    ngStorage       : "bower_components/ngstorage/ngStorage.min.js",
+    sortable        : "js/sortable.js",
+    angularPage     : "js/angular-page.js",
+    mainControllers : "js/main/main-controllers.js",
+    searchControllers: "js/main/search-controller.js",
+    mainService     : "js/main/main-service.js",
+    mainDirective   : "js/main/main-directives.js",
+    appConfig       : "js/app-config.js",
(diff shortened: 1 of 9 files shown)


nblair:
It looks like the 'my-app.js' used by the app model (see UW-Madison-DoIT/my-app-seed) is supplanted by main.js. Can you show how my-app-seed will need to be retrofitted?

angularjs-portal-frame/src/main/webapp/main.js
```diff
@@ -0,0 +1,59 @@
+require.config({
+
+  paths: {
+    'my-app': 'my-app.js'
+    /* TODO: Incorporate the <scripts/> that are being globally imported in frame.html
```
timlevett: What is wrong with just doing this now?

paulerickson: Happy to - and I started to - but I was biting off a bit too much at once (the bower config could also use TLC), so I thought it was better to start with the minimum functional change and circle back to get things better organized.

paulerickson:
I'll check out my-app-seed tomorrow and make the change.  my-app.js is still the extension point for app code, only now being included by main.js rather than frame.html directly.  my-app.js could be called directly as it is, but using main.js gives a place to manage the shared libraries, initialization, or any other config that shouldn't be overlaid by the app.

angularjs-portal-frame/src/main/webapp/main.js
nblair: I'm with @paulerickson here, small bites :)

I think that can come with the next iteration.

timlevett: I'm just curious what you ran into that made this hard.

paulerickson: I'd convert our own code to require modules, and most of the libraries need shim configurations, which just means going through one by one to find what they depend on and what they export.  It's mostly rote, but there might need to be changes in how things initialize, like I had to do here for angular.  I've heard that Google Analytics can break inside the async loading of require.js without some change, but I didn't dig in to see how much effort is needed.

timlevett: Ah, i see. Yes then I agree to wait.

nblair:
The changes here in conjunction with https://github.com/UW-Madison-DoIT/my-app-seed/pull/2 should allow us to proceed. Has :+1: from me.

PR #186, opened 2015-04-07
MUMUP-1789 : Update to public releases

paulerickson:
+ Remove parent so we don't distribute to uw-releases
+ Minor fix to URL for scm block
+ Update repository to public-releases

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,11 +1,5 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-  
-  <parent>
-    <groupId>edu.wisc.doit.code</groupId>
-    <artifactId>shared-tools-uw-parent</artifactId>
-    <version>9</version>
-  </parent>
-  
+
   <modelVersion>4.0.0</modelVersion>
   <groupId>edu.wisc.my.portal.angular</groupId>
   <artifactId>angularjs-portal-parent</artifactId>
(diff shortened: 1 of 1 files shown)


timlevett:
If we are going to remove the parent pom to distribute to uw-releases then we should add in `<distributionManagement>`

See https://maven.apache.org/pom.html#Distribution_Management for details.

nblair:
Well, why not Maven Central? This project's source is fully public, and I'm not sure we really truly want to continue having a local, publicly accessible, artifact repo.

paulerickson:
I quite agree!  I ran it by Tim in chat and got no objections, so I cut a ticket for Sonatype to create the project: https://issues.sonatype.org/browse/OSSRH-14879

timlevett:
Dig it.

paulerickson:
Oh I almost forgot, we'll need our build-box administrator (@nblair?) to update .m2/settings.xml with server credentials and a gpg passphrase, if not already there.

The authorized uploaders right now are levett, paulerickson, nblair, jhanstra

The configuration would look something like:
```xml
        <profiles>
                <profile>
                        <id>ossrh</id>
                        <activation>
                                <activeByDefault>true</activeByDefault>
                        </activation>
                        <properties>
                                <gpg.executable>gpg2</gpg.executable>
                                <gpg.passphrase>the-gpg-passphrase</gpg.passphrase>
                        </properties>
                </profile>
        </profiles>
        <servers>
                <server>
                        <id>ossrh</id>
                        <username>paulerickson</username>
                        <password>the-sonatype-password</password>
                </server>
        </servers>
```

timlevett:
Hmmm, i think this should probably run on the my-tools build box. Does that make sense to you guys? I'm curious if we need to create a myuw-bot sonatype account. I would like to not add my creds to a build server

vertein:
I tend to avoid putting my credentials on shared things.  

Also, I wouldn't mind be authorized to upload this project as well.

timlevett:
I'm also going to go ahead and delete the .19-SNAPSHOT from code.doit/uw-snaps

paulerickson:
I wasn't aware of the my-tools box; could I get access?  Is that where uw-ui-toolkit is set up?  That project is in Central as well, so maybe they could use the same credentials.

Maven can at least obscure plaintext passwords with `mvn --encrypt-password`, which generates a hashed-up version, although it remains a shared credential that you can eff with if you can run maven on the box.

timlevett:
:+1:

PR #188, opened 2015-04-08
MUMUP-1816: Removed serlet from WAR, updated parent pom for updated repositories.

tzodrow:
Updated WAR pom to remove unneeded dependencies - blocking development and not needed here
Updated Parent pom to update repository information

### DIFF
--- a/angularjs-portal-frame/pom.xml
+++ b/angularjs-portal-frame/pom.xml
@@ -13,23 +13,6 @@
   <name>angularjs-portal-frame</name>
   
   <url>https://github.com/UW-Madison-DoIT/angularjs-portal</url>
-  
-  <dependencies>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>3.8.1</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-        <groupId>javax.servlet.jsp.jstl</groupId>
-        <artifactId>jstl-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>jstl</groupId>
-      <artifactId>jstl</artifactId>
-    </dependency>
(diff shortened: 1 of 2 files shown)


nblair:
:+1:

paulerickson:
:+1:

timlevett:
@nblair @paulerickson @tzodrow  - this commit broke /frame, it is now undeployable. Was this tested?

timlevett:
we are using jstl core in index.jsp

nblair:
We tested with our application, but testing with MUM is not a known quantity to us. Is the dependency needed in -home and not -frame? What artifact are you deploying?

nblair:
Let's get travis-ci enabled for this project, and come up with some builds that can help verify.

timlevett:
We have frame setup as an independent app. Just /frame doesn't work.

timlevett:
It builds fine, but when you hit index.jsp it throws an error because it can't find `http://java.sun.com/jsp/jstl/core`

paulerickson:
I guess I've never hit index.jsp before, so I never saw the JSP compiler complain.  Is index.jsp doing anything?  It includes frame.html, but I don't see any jstl tags there.

timlevett:
The `frame.html` has a versioned CSS file with the project version. Instead of changing that everytime we want to change the release we are using `${project.version}`. I'm all for alternative methods if you have a suggestion, this was just how we knew it would work. If that makes sense.

I'm surprised your styling isn't messed up if you didn't hit index.jsp. Are you just using frame.html? Or are you overwriting the index page all together?

paulerickson:
Ohh, I see.  I did run into that intermittently, actually, but I blamed it on my local changes with requirejs.  I have to think my IDE is compiling the JSP despite the missing pom dependency or something.

PR #190, opened 2015-04-09
MUMMNG-1305: Change default smart widget call-to-action to sentence case

jhanstra:
image

### DIFF
--- a/angularjs-portal-home/src/main/webapp/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/partials/widget-card.html
@@ -49,7 +49,7 @@
     <div class="widget-icon-container">
       <portlet-icon></portlet-icon>
     </div>
-    <button class="btn btn-default launch-app-button">Launch Full App</button>
+    <button class="btn btn-default launch-app-button">Launch full app</button>
   </a>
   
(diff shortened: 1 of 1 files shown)


vertein:
:+1:

apetro:
:+1: as far as it goes.  

@timlevett had been suggesting this might be an opportunity to refactor the call-to-action button to be a directive, maybe one that takes a parameter of the call-to-action text and defaults when that parameter not provided, so that we've abstracted out this control and can adjust it in one place affecting all its usages in the future.  Thoughts on whether that's worth doing at this juncture?  `#refactor_mercilessly`.

keirserrie:
:+1:

jhanstra:
@apetro @timlevett I'm not totally sold yet that there's a huge amount of value to be gained from making the button a directive - and the way our code is currently set up, it will require us to make some decisions about touch/click surfaces (because some of those buttons are `<button>`s and some are `<a>` links - in basic widgets the whole box is a link currently). So I think it's out of scope right now, but let's pick up a conversation about it when Tim gets back and see if it's worth doing.

timlevett:
I agree @jhanstra . Might be challenging to create one.

PR #191, opened 2015-04-10
MUMMNG-1304 Adds tooltip to widget selector

vertein:
Gif looks a bit choppy, but it's the same settings as the tooltip on the '+' button.

viewtooltip


apetro:
:+1:

keirserrie:
Looks like space being added after "Simple" messing up centering

angularjs-portal-home/src/main/webapp/partials/home-toggle.html
```diff
@@ -1,4 +1,23 @@
 <span class="layout-chooser" ng-if="toggle">
-  <li ng-class="{active : modeIs('list')}"><a  href="javascript:;" ng-click="switchMode('list')"><i class="fa fa-th"></i></a></li>
-  <li ng-class="{active : modeIs('widgets')}" ><a href="javascript:;" ng-click="switchMode('widgets')"><i class="fa fa-th-large"></i></a></li>
-</span>
\ No newline at end of file
+  <li ng-class="{active : modeIs('list')}">
+    <a  href="javascript:;" ng-click="switchMode('list')"
```
apetro: I suspect that we ought to be doing something to this `<a` tag to improve its accessibility characteristics.  As in, a `title` or `ARIA`-something-or-other or a screen-reader-only alternative content for the tag?  Link text only visible to screen readers might be brilliant here, in that if I can't see the `fa-th`, what I really want is the text equivalent.  Maybe that should be happening on the `<i` tag?

jhanstra:
Not sure if we've talked about it before, but I kind of like 'Content View' or 'Widget View' more than 'Detail View'...thoughts?

keirserrie:
I could go for Widget View...others have thoughts?

vertein:
Right about the spacing thing, corrected.  Widget View sounds good to me.

apetro:
I think almost any text trying to articulate a text version of this is better than no text, so :+1: for progress.

I suspect that we ought to be getting to some specific, consistent terms of art, names for the features in MyUW that we need to be able to talk about and that users need to be able to talk with us about.  So, eventually, this is "Smart Widgets" vs "Minimal Widgets", and we're quite specific about what we call these, and whether those are proper nouns or "smart widgets" vs "minimal widgets", etc.  

I think we should take whatever steps towards this we can in this story, and if we don't get clear enough about this, we're going to end up revisiting it as the need becomes clearer when KB etc. documentation catches up to talk more about Beta features.

keirserrie:
What do you guys think, does labeling "smart widgets" and "simple widgets" make sense to a user?

timlevett:
:+1:

vertein:
The decision is to now use 'icons' and 'widgets'.  I will update and merge the pr shortly

vertein:
Changed the wording to 'icons' and 'widget's

keirserrie:
Sorry...can you capitalize Icons and Widgets. Going w/sentence case for most things and want to be consistent.

jhanstra:
@keirserrie, I went in and capitalized it and pushed a commit to master.

keirserrie:
@jhanstra THX

PR #192, opened 2015-04-10
Add my-app.css so app developers can style their apps

jhanstra:
Eventually might want to make this a `less` file, but css to start. This is the same idea as the RequireJS work @paulerickson did recently. Give app devs a file to style anything they need to within their app.


paulerickson:
:+1: 
RequireJS can bring in CSS too, but their own [documentation](http://requirejs.org/docs/faq-advanced.html#css) doesn't make it sound super compelling

vertein:
:+1:

PR #194, opened 2015-04-15
Initial travis-ci configuration

nblair:
Looks like we are all setup already on travis-ci.org, we just need the config file for builds to start.


timlevett:
:+1:

PR #196, opened 2015-04-15
Add back in jslt dependency

jhanstra:
So /frame works.

image


timlevett:
Opps, looks like this is a duplicate on https://github.com/UW-Madison-DoIT/angularjs-portal/pull/197. I would rather do #197 as it changes less and has the same result.

timlevett:
Closing as a dup

jhanstra:
Sounds good

PR #199, opened 2015-04-16
[MUMUP-1820] Finish incorporating RequireJS for better maintainability & encapsulation

paulerickson:
> I tried to avoid changing the guts of scripts, but converting to RequireJS modules changed the indent level here and there, so I suggest adding `&w=1` to the URL to ignore whitespace.

## Goals ##
1. **Clean up the global namespace**
   * Explicitly declaring dependencies helps protect frame & app writers from each other, and makes library upgrades / deprecations more manageable
2. **Organize resources into modules** so that it's
   * Easy to find angular code and partials
   * Easy to move things around later
   * Hard to accidentally overlay frame content (e.g. partials/settings.html) in downstream projects

## Changes ##
* **Our existing scripts are decorated with Require module definitions**, which is an extension of the _immediately executed factory function_ pattern we already use...
  * i.e., instead of `(function() {...})()`, dependencies are declared by name so that RequireJS can provide them: `define(['<dependency>'], function(dependency) {...})`
* **Other scripts are shimmed into RequireJS
* **Files are organized after the angular modules they define**, for instance:
  * The angular module `portal.misc.directives` is in `portal/misc/directives.js`
  * I renamed `angular-page.js` to `portal.js`, to match the name of its angular module
    * I renamed it again to `portal/main.js` to make it a RequireJS [package](http://requirejs.org/docs/api.html#packages) named `portal`, which contains other modules.
  * Paths to partials are define relative to the modules that reference them - this ensures that paths will not overlap, and modules can be moved around with breaking paths.

## TODO ##
* This isn't merge-ready yet, because I only touched angularjs-portal-home enough to break it, but I wanted some feedback on -frame before continuing.  **Thoughts?**
* Version should be updated, since removing `<script>` imports is essentially backwards-incompatible for consumers.


timlevett:
Really liking the restructure. Very clean. :+1: for continuing.

angularjs-portal-frame/src/main/webapp/frame.html
```diff
@@ -53,26 +53,7 @@
...
-<script type="text/javascript" src="js/main/main-directives.js"></script>
-<script type="text/javascript" src="js/app-config.js"></script>
-<script type="text/javascript" src="js/controllers.js"></script>
-<script type="text/javascript" src="js/services.js"></script>
-<script type="text/javascript" src="js/directives.js"></script>
-<script type="text/javascript" src="js/filters.js"></script>
-<script type="text/javascript" src="js/ui-bootstrap.js"></script>
-<script type="text/javascript" src="js/ga.js"></script>
-<script type="text/javascript" src="bower_components/requirejs/require.js" data-main="main.js"></script>
+  <script type="text/javascript" src="js/ga.js"></script>
```
paulerickson: The thing about ga.js is that it already does its own asynchronous loading, so putting it inside the asynchronous loader of require.js is redundant / requires initialization configuration.
Because it's mainly being called from markup rather than other scripts, and it's the only one in this situation, I figured there's nothing really to be gained by moving it to Require.

timlevett: I agree with this logic.

paulerickson:
Shouldn't angularjs-portal-home overlay `my-app` to add its content, in the same way we are doing with my-student-financial-account?

timlevett:
@paulerickson - `angularjs-portal-home` overlays on `angularjs-portal-frame` just like the app seed does. If you are overlaying on the app-seed that is weird, but to each there own. From my understanding, the app-seed is supposed to be a starter, not an overlay.

timlevett:
Minor correction. They are not exactly the same. Currently home overwrites things, but we haven't had time to clean it up.

paulerickson:
@timlevett I should back up.  `angularjs-portal-frame` provides this stub of a file (was `my-app.js`, now `my-app/main.js`) that is included in the root `main.js` (was in `frame.html`).  `my-app-seed` and `my-student-financial-account` overlay that as the entry point for application stuff.  `angularjs-portal-home` leaves my-app and _instead_ overlays the 'portal' module (was `angular-page.js`, now `portal/main.js`).  So, I'm thinking it should leave _that_ alone and actually put its contents in my-app; I just want to make sure I'm not missing something about why it's laid out this way.

timlevett:
Yeah, that makes more sense @paulerickson .

paulerickson:
Cool, thanks

angularjs-portal-frame/src/main/webapp/frame.html
jhanstra: Sounds good

paulerickson:
Alright, there are still some encapsulation concerns, but I think they are at least in the angular side of the court now.

Now, `angularjs-portal-home` is extending from / overlaying the my-app package instead of portal (frame).  Since the app replaces whatever routeProvider config that was in the base/frame project anyway, I moved that config into my-app and made my-app the thing that angular bootstraps.  So there's still a question of how one might handle pages/routes on the frame itself, preferably without needing each individual app to wire them up.
I also extracted the route values out of the route configuration, along with their partials, so that they can reside in their corresponding modules and use relative paths.

angularjs-portal-home/src/main/webapp/my-app/main.js
```diff
@@ -0,0 +1,68 @@
...
+            when('/apps/details/:fname', marketplaceRoutes.details).
+            when('/apps/search/:initFilter', marketplaceRoutes.main).
+            when('/features', featuresRoute).
+            when('/list', listRoute).
+            when('/notifications', notificationsRoute).
+            when('/settings', settingsRoute).
+            when('/static/:fname', staticRoute).
+            when('/widgets', widgetRoute).
+            otherwise({ redirectTo : '/list'});
+    }]);
```
timlevett: This approach is neat. Kind of agree with the todo, but this is a good start.

timlevett: Thinking about it, I think Google did a presentation on how they did this with there cloud service. They basically created a master list with lazy loading.

timlevett:
Brought it down locally against the latest master (with the mock portal) and found a couple of things.

+ ui.sortable is complaining about jquery && angular. Error message: `ui.sortable: jQuery should be included before AngularJS!`. This breaks sorting on `/lists` and `/widgets`
+ directives.js needs jquery for the .hide() method.
+ This seems to have reverted the html5 mode flag getting set that #200 added. 

Overall looks like we are almost there.

Woot!

paulerickson:
Yikes, thanks for catching those, @timlevett !  I think I misplaced html5Mode with the last merge.

Re: the first two bullets, I think this is because `angular.js` itself uses jquery under the hood.  I turns out it has its own implementation, JQLite, that it can use when it doesn't have the real thing.  #TIL

jhanstra:
This is fantastic, great job! Agree with the things @timlevett found.

paulerickson:
There's still one error I get:
```
TypeError: undefined is not a function
  at app.value.directive.link (http://localhost:8080/web/js/sortable.js:310:23)
```

I thought this was missing data that would go away with the addition of the mock portal, but no such luck.

timlevett:
:+1: Looks good, :shipit:

PR #200, opened 2015-04-16
MUMUP-1799: Hashless Links

jhanstra:
So, I got HTML 5 mode to work. But we will need a server side component to this.

When you go to /web, you can go to /web/apps by clicking the sidebar, and angular will change the route/url for you. But if you refresh on that page, you get a 404 because the server doesn't know what to do with that. (Sidenote: I don't really understand how the server knows what to do with /web/#/apps though...is that this html5 mode magic?)

I will definitely need help with the server side part. @timlevett @mjadas do you know how we could do this?

screen recording 2015-04-16 at 10 45 am

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/frame.html
+++ b/angularjs-portal-frame/src/main/webapp/frame.html
@@ -53,7 +53,6 @@
 
   <!-- FOOTER  -->
   <site-footer></site-footer>
-  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.3.15/angular.js"></script>
 <script type="text/javascript" src="js/config.js"></script>
 <script type="text/javascript" src="bower_components/jquery/dist/jquery.min.js"></script>
 <script type="text/javascript" src="bower_components/jquery-ui/jquery-ui.min.js"></script>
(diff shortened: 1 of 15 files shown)


angularjs-portal-frame/src/main/webapp/frame.html
```diff
@@ -53,6 +53,7 @@
 
   <!-- FOOTER  -->
   <site-footer></site-footer>
+  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.3.15/angular.js"></script>
```
timlevett: Don't we already have `angular.js` in the project?

jhanstra: Oops, from troubleshooting

timlevett:
I can help you out with the server side part @jhanstra

timlevett:
Added in server side stuff. :smirk_cat:  Kind of stinks we have to list out all the subdirectories, but you can't do just `/*` as that creates infinite redirects and server goes boom :crying_cat_face:

jhanstra:
Thanks @timlevett. /apps and /settings and /static all work, but marketplace details do not. I think I know how to fix it so I'll take a stab at it.

timlevett:
Urhhhm. details works for me. Example url: http://localhost:8080/web/apps/details/library-reserves-by-department

jhanstra:
Got it. Just incorrectly changed my links locally, whoops. We're good to merge now if we get a +1.

timlevett:
:+1:

timlevett:
Note we should update the links in header and side bar on /portal as well.

PR #203, opened 2015-04-17
[MUMUP-1822] Allow use of the jetty plugin AND contexts other than "/frame/"

paulerickson:
Phew, I just wanted to fix the issue of missing `angular.${project.version}.css` when running with the jetty plugin, but things unraveled a bit as I yanked that thread.

1.  `angularjs-portal-frame` couldn't be run with `jetty:run` because it looks in the source directory, but `angular.${project.version}.css` winds up in the target directory.
  * :white_check_mark: Configured jetty to look there
2.  `angularjs-portal-home` had an undeclared tomcat dependency :frowning: 
  * :white_check_mark: Declared dependency as provided (and provided from within the jetty plugin)
3. [Hashless links](https://github.com/UW-Madison-DoIT/angularjs-portal/pull/200) made sexy URLs, but hardcoded the context as "/frame/".  This is inconvenient if you're running an individual app from jetty, or if you don't want your final app to be named "frame".
  * :white_check_mark: Set the `<base>` depending on the _actual_ rather than assumed context path.  I made the page a JSP to be able to access this.

I made sure I could run the frame and home submodules individually from jetty ("/" root context), from the parent ("/frame/" and "/web/"), and dropped into Tomcat 7 container.

This should make it easier to develop an app in isolation, and run outside of the shared MUM/portal tomcat container.


angularjs-portal-frame/pom.xml
```diff
@@ -38,33 +42,46 @@
...
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-maven-plugin</artifactId>
+        <version>9.3.0.M2</version>
+        <configuration>
+          <webApp>
+            <baseResource implementation="org.eclipse.jetty.util.resource.ResourceCollection">
+              <resourcesAsCSV>src/main/webapp,${webapp.target.directory}</resourcesAsCSV>
```
paulerickson: This works a lot like war overlays - when a request is made, it looks in src first and then the target dir.  Of course, changes in the .less file won't be recompiled automatically.

angularjs-portal-frame/src/main/webapp/index.jsp
```diff
@@ -1,2 +1,3 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
-<c:import url="/WEB-INF/classes/frame.html" />
+<jsp:include page="/WEB-INF/classes/frame.jsp" />
```
timlevett: So we _could_ just not have 2 JSPs at this point, and move the content of `frame.jsp` to `index.jsp`.  Thoughts?

angularjs-portal-frame/pom.xml
timlevett: Interesting. #TIL

jhanstra:
Awesome, :+1:. I was so happy I got HTML5 mode to work until I realized it broke *everything* for app devs. Nice fix.

angularjs-portal-home/pom.xml
```diff
@@ -41,39 +41,45 @@
...
     <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
       <scope>provided</scope>
     </dependency>
-    
+
+    <dependency>
+      <groupId>org.apache.tomcat</groupId>
+      <artifactId>tomcat-catalina</artifactId>
```
paulerickson: web.xml uses ExpiresFilter from this

timlevett:
:+1: Nice improvements! One minor optional suggestion

angularjs-portal-home/pom.xml
timlevett: True story, good catch.

angularjs-portal-frame/pom.xml
paulerickson: Yeah, I never even realized jetty:run looks at source rather than target.  Maybe `mvn jetty:run-war` is the sensible thing to do, but this is faster.

angularjs-portal-frame/src/main/webapp/index.jsp
timlevett: Talked in chat, this is needed so `jetty:run` is happy

PR #205, opened 2015-04-17
MUMUP-1824 : add mock portal env

timlevett:
This idea emerged after @paulerickson added the jetty maven plugin to `angularjs-portal`.

To run: `mvn jetty:run`

Adds `/portal` with some static JSON. This could be better, but at least you can view everything right now. Note add/edit/remove all don't work properly.

Cheers,

Tim


paulerickson:
Awesome!  :+1:

jhanstra:
:+1:

PR #208, opened 2015-04-28
MUMUP-1818 : move widget template in layout feed

timlevett:
Relates to https://github.com/UW-Madison-DoIT/uPortal/pull/352

+ Moves where to get the template
+ Add in error if the template is expected but not present


jhanstra:
:+1:

PR #209, opened 2015-04-28
MUMUP-1814: Bug: weather  widget "World Weather Online" link doesn't open in new window

Daha5:
The world weather online link at the bottom of the weather widget doesn't open in a new tab when clicked. I added target="_blank" to the HTML code and now it works. 

For reference, the link this issue is about is at the bottom of this screenshot: "Weather provided...
linkatbottom


timlevett:
:+1:

timlevett:
Fixed in #210

PR #211, opened 2015-04-29
Add in ability to reuse loading icon

timlevett:
+ Adds the `reuse` attribute to the loading icon
+ Adds example to the /settings page for reuse and one time use

http://goo.gl/KcMeQ5

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/portal/misc/directives.js
+++ b/angularjs-portal-frame/src/main/webapp/portal/misc/directives.js
@@ -8,6 +8,7 @@ define(['angular', 'require'], function(angular, require) {
      * REQUIRED attribute that isn't listed below:
      *   object : this is the scope array we are watching to show/hide gif
      *   empty  : this is the scope boolean flag that you set if the data came back and it was empty
+     *   reuse  : (optional) set to true, it won't destroy the loading gif, just hide it
      *
      */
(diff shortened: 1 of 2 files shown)


vertein:
I like it.  Good thinkin' :+1:

angularjs-portal-frame/src/main/webapp/portal/misc/directives.js
```diff
@@ -8,6 +8,7 @@ define(['angular', 'require'], function(angular, require) {
      * REQUIRED attribute that isn't listed below:
```
apetro: I don't get what "attribute that isn't listed below" means, now that I read this.  (Acknowledging that that wasn't added by this changeset. :smile:)

angularjs-portal-frame/src/main/webapp/portal/misc/directives.js
```diff
@@ -8,6 +8,7 @@ define(['angular', 'require'], function(angular, require) {
...
      *   object : this is the scope array we are watching to show/hide gif
      *   empty  : this is the scope boolean flag that you set if the data came back and it was empty
+     *   reuse  : (optional) set to true, it won't destroy the loading gif, just hide it
```
apetro: Suggestion : be clearer that defaults to `false`.

PR #213, opened 2015-05-04
MUMMNG-1386: Messaging for widgets

jhanstra:
image

I'm not sure how this works across environments. Like, where is the new-stuff.json coming from in production? I may have implemented this wrong if there is a separate feed for each environment.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
@@ -409,7 +409,7 @@ div.table-cell {
     position:relative;
     top:3px;
     font-size:12px;
-    left:2px;
+    left:5px;
   }
 }
 
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
@@ -400,6 +400,18 @@ div.table-cell {
   width:100%;
   max-width:100%;
 }
+.header-opaque .second-description {
+  float:right;
+  p {
+    font-weight:600;  }
+  span {
+    font-weight:600;
+    position:relative;
+    top:3px;
+    font-size:12px;
+    left:2px;
+  }
+}
 
 @demo-color:@color1;
 @demo-color-darker:@color2;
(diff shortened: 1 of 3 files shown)


vertein:
:+1: @jhanstra promised me that on screen size changes, it still works.

angularjs-portal-home/src/main/webapp/samples/new-stuff.json
```diff
@@ -2,11 +2,11 @@
             [
               { 
                 "id" : 1,
-                "expireYr"  : 2016,
-                "expireMon" : 2,
-                "expireDay" : 20,
-                "title" : "AngularJS Portal",
-                "shortDesc" : "Welcome to the alternative home page to uPortal.",
+                "expireYr"  : 2015,
+                "expireMon" : 5,
```
timlevett: Month is 0 based, fwiw.

http://www.w3schools.com/jsref/jsref_obj_date.asp

timlevett:
New stuff is in the overlay for some environments. So just flag this issue with label mum-overlay.

I'm not 100% about the second desc hard coding. If we do accept this please create an issue to refactor/remove it later.

overall, :+1:

angularjs-portal-home/src/main/webapp/samples/new-stuff.json
jhanstra: Ah yes, good catch.

apetro: Is there some reasonable way to comment this trip hazard?

angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
```diff
@@ -400,6 +400,18 @@ div.table-cell {
   width:100%;
   max-width:100%;
 }
+.header-opaque .second-description {
```
apetro: Is this "second-description", or is this maybe "smart-widget-selector-message" ?  Seems like semantically this is less about being a second description and more about being the specific magic incantation required to get that pointer to float pointing at that favorite mode selector control.

apetro:
:+1:

jhanstra:
@timlevett I removed the hard-coded code so we don't need to remember to take it out. It's kind of janky as it adds one-off attributes to the json feed but I don't think that's a huge deal. Let me know what you think.

vertein:
I think I like it better in the JSON.  What happens when that second description isn't there?

jhanstra:
@vertein it should just not show anything.

vertein:
Cool, just checking. :)

timlevett:
:+1:

PR #214, opened 2015-05-05
MUMMNG-1396: Weather Widget should link to UW-Madison weather source

Daha5:
If the user has no weather locations set, then the widget will provide a link to the UW-Madison institutional weather source (http://www.ssec.wisc.edu/localweather/).


vertein:
:+1:

apetro:
:sunny:

PR #215, opened 2015-05-05
[SPFP-81]  Redirect to index by default

paulerickson:
Having to take every URL pattern from your angular app and copy it into servlet-mappings is a pain!
Moreso if you wanted to use Java based config in your app (SPFP), because XML configuration gets a higher precedence than annotation or Java-based config.

This change sends any would-be 404 to the angular app, although clients like `angularjs-portal-home` can still overlay web.xml to get that fine-grained control and explicitly map as currently.

Example:
```
/frame/ --> default servlet --> index.jsp --> angular $routeProvider --> main page
/frame/settings --> default servlet --> not found --> index.jsp --> angular $routeProvider --> settings page
/frame/quuux --> default servlet --> not found --> index.jsp --> angular $routeProvider --> main page
/web/quuux --> default servlet --> not found --> 404
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,14 +17,14 @@
         <name>My UW Dev Team</name>
     </developer>
   </developers>
-  
+
   <licenses>
     <license>
       <name>The Apache License, Version 2.0</name>
       <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
     </license>
   </licenses>
-  
+
   <scm>
     <connection>scm:git:git@github.com:UW-Madison-DoIT/angularjs-portal.git</connection>
     <developerConnection>scm:git:git@github.com:UW-Madison-DoIT/angularjs-portal.git</developerConnection>
@@ -165,6 +165,7 @@
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-maven-plugin</artifactId>
         <version>9.3.0.M2</version>
+        <inherited>false</inherited>
         <configuration>
           <contextHandlers>
             <contextHandler implementation="org.eclipse.jetty.maven.plugin.JettyWebAppContext">
--- a/angularjs-portal-frame/src/main/webapp/WEB-INF/web.xml
+++ b/angularjs-portal-frame/src/main/webapp/WEB-INF/web.xml
@@ -2,7 +2,7 @@
  "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN"
  "http://java.sun.com/dtd/web-app_2_3.dtd" >
 
-<web-app 
+<web-app
     xmlns="http://java.sun.com/xml/ns/javaee"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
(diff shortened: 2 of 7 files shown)


apetro:
Clever.  Would it be feasible to nuance this to:

```
/frame/quuux --> default servlet --> not found --> index.jsp --> angular $routeProvider --> a client-side 404-style page
```
?

As in, some kind of default client-side 404 not found experience when the angular routeProvider, well, doesn't find a route for a path?

timlevett:
@apetro - true you could do that in your route provider for the otherwise clause.

paulerickson:
Yes!  In my opinion, that should vary by app, and there might still be more useful behaviors than a 404 page - i.e. going to the main page seems sensible for a one-tab site like /frame, but for /web maybe it should route it to search (maybe I type /apps/whether, and the search interface helps me along to /apps/weather)?  For bucky-backup, I might just assume the path was a node ID, and show a node-not-found error if need be.

pom.xml
```diff
@@ -165,6 +165,7 @@
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-maven-plugin</artifactId>
         <version>9.3.0.M2</version>
+        <inherited>false</inherited>
```
paulerickson: This is unrelated, but gets rid of errors when running jetty:run from the subprojects directly because it would pick up the parent configuration.

timlevett:
So I think we decided best course of action would be to make the web.xml as minimal as possible. So commenting out the mapping for `/settings`. We welcome all thoughts.

paulerickson:
Alright, the mapping is out, the duplication of client & server side routing is opt-in rather than opt-out, consumers can start building apps without changing server-side code, and I added a basic error page (nicked from pull/219) that can be used as-is or as a template for app-specific handling.

not-found

timlevett:
Hmmm, I still see the redirects for 404 to index.jsp and the setting static hack in the JSP. I would prefer not do those things. thoughts?

paulerickson:
Sure, the intent with the 404 -> index.jsp mapping is that any requests the server doesn't know how to handle are delegated to the angular app.  I see now that would NOT be so good if you want to package webservices in the same war, which is just what sunblock does :frowning:.  I think it's best to keep services and UIs in separate wars and separate JVMs for the sake of security and scalability... but that's another issue I guess.
Is this the concern, or are there other scenarios I'm not thinking of?

What if we take out the 404 bit and make servlet-mappings? (I know we discussed some of this, but I don't remember if we settled on something particular)
```
/api/* --> some new default API servlet
/* --> index.jsp
```
or
```
/app/* --> index.jsp
```

paulerickson:
The more I think about it (too much), the more I don't like having a REST endpoint alongside this single-page-app -- at least with html5Mode, because it makes the requests look like resources instead of a single page.
What I'm thinking would be elegant is all requests for `text/html` go to `index.jsp`, and all `application/json|xml|etc` is the business of whatever custom code and endpoints consumers define.  Thoughts?

paulerickson:
Okay, made some progress with a content negotiation approach, but it's enough configuration and potential for issues that I'm throwing in the towel.  I'll stash my changes and hopefully tackle this another time, but for now I'll overlay web.xml in my-student-financial-account to keep things moving.

apetro:
@paulerickson I think that's the right tactical outcome.  These are worthy problems and I really appreciate your digging into this, but I think one thing to realize out of this is that we're just not ready to solve this problem yet, and there's not enough pressure yet.  Present-generation settings path just isn't that important.

This problem domain will no doubt resurface.  Please do feel free to keep your branch / stashed changes up somewhere public so it'll be handy on the resurfacing.

paulerickson:
Thanks, @apetro.  Just to clarify, though, the main problem to solve IMO is that the `servlet-mapping` will inevitably get out of sync with `$routeProvider`, and result in bugs and regressions, since it can be subtle enough that you might not notice in testing unless you're looking for it.

I'll leave this Spring content-negotiation tutorial here for reference on that, which will be more useful than the mess in my git stash:
https://spring.io/blog/2013/06/03/content-negotiation-using-views

PR #217, opened 2015-05-05
MUMMNG-1393: Google Analytics for smart widgets

jhanstra:
Triggers an event on click of both icon tabs - icon view and widget view.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-toggle.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-toggle.html
@@ -5,6 +5,7 @@
       tooltip-trigger="mouseenter"
       tooltip-placement="top"
       tooltip-popup-delay="500"
+      onClick="_gaq.push(['_trackEvent', 'Widgets', 'List Mode', 'List Mode']);"
     >
         <i class="fa fa-th"></i>
     </a>
@@ -15,6 +16,7 @@
       tooltip-trigger="mouseenter"
       tooltip-placement="top"
       tooltip-popup-delay="500"
+      onClick="_gaq.push(['_trackEvent', 'Widgets', 'Widget Mode', 'Widget Mode']);"
     >
       <i class="fa fa-th-large"></i>
     </a>
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-toggle.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-toggle.html
@@ -5,7 +5,6 @@
       tooltip-trigger="mouseenter"
       tooltip-placement="top"
       tooltip-popup-delay="500"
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-toggle.html
```diff
@@ -5,6 +5,7 @@
       tooltip-trigger="mouseenter"
       tooltip-placement="top"
       tooltip-popup-delay="500"
+      onClick="_gaq.push(['_trackEvent', 'Widgets', 'List Mode', 'List Mode']);"
```
timlevett: I think it would be better to just add a call to [miscService.pushGAEvent](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-frame/src/main/webapp/portal/misc/services.js#L29) in the `switchMode` function.

apetro: Hmm.  I think the API is `_trackEvent, {category}, {action}, {label}, {value}` .  So I'd read this as a "Widgets" category event with action "List Mode" and label "List Mode"?

Would it be better to go for something more like  

```
onClick="_gaq.push(['_trackEvent', 'Widgets', 'mode-select', 'List Mode']);"
```

so that the widget mode selection would have the same `action` so that these can be compared by label?  I don't know that it makes much difference, just trying to think through how to use the API to organize the taxonomy of events.

timlevett: I agree with @apetro's comment

jhanstra: Oh interesting, that service is pretty easy. I'll use that. I was going off of the Beta event in beta-header.html:
`<a onClick="_gaq.push(['_trackEvent', 'Beta', 'Read About Change', 'Read About Change']);" 
	          	   href="http://redesign.my.wisc.edu/" 
	          	   target="_blank">
	          		Read about what's changing
	          	</a>.`
Should I refactor this as well?

timlevett: Hmmmmmmmmm, maybe? Probably would be best just to create a Jira.

apetro: Backlogged a more holistic refactor of Google Analytics events usage : https://jira.doit.wisc.edu/jira/browse/MUMUP-1839 .

timlevett:
:+1: :golf: :clap:

timlevett:
Thought, maybe add some documentation about what this means in https://wiki.doit.wisc.edu/confluence/display/MUM/Google+Analytics+in+MyUW

PR #218, opened 2015-05-06
Fix details link overlap

jhanstra:
Phyllis pointed out in the sprint review that for apps without a category, the Details link overlaps the bottom border. After fixing this I couldn't find an example in production so I don't think this is actually an issue...but now it's fixed just in case :)

Localhost before:
image

After:
image


vertein:
:+1:

PR #219, opened 2015-05-06
MUMUP-1428: Error experience for <IE10 users

jhanstra:
So Angular straight up doesn't support IE 8, but I don't think our experience on IE 9 is that great either. Do we want to support IE 9 or leave it in the dust? Right now I have it not supported, but it's easy to change it to support IE9.

This code will display a simple div at the top of the page informing users browsing on < IE 10 that they are using an outdated browser and they should update it. This snippet is commonly used on many other websites and links to a site called Browse Happy that shows users how to download a better, up-to-date browser. Personally, I think this would be more engaging and intuitive than linking to a KB doc, but I understand if we want to be more UW-specific with this and go that route instead. Thoughts?

Screenshot of how IE8 would look (I don't have an easy way to test on IE9):
image

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
@@ -451,9 +451,14 @@ div.table-cell {
 }
 
 .browserupgrade {
+  position:fixed;
+  height:100%;
+  width:100%;
+  background-color:#fff;
   text-align:center;
-  padding:20px;
+  padding:100px;
   font-size:18px;
+  z-index:9000;
   span {
     font-size:50px;
     margin:30px 0px 10px;
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/home.less
@@ -450,6 +450,20 @@ div.table-cell {
   z-index:999;
 }
 
+.browserupgrade {
+  text-align:center;
+  padding:20px;
+  font-size:18px;
+  span {
+    font-size:50px;
+    margin:30px 0px 10px;
+  }
+  a {
+    color:@color1;
+    font-weight:600;
+  }
+}
+
 
 @media (max-width: 767px) {
   div.table {
--- a/angularjs-portal-frame/src/main/webapp/frame.jsp
+++ b/angularjs-portal-frame/src/main/webapp/frame.jsp
@@ -21,7 +21,7 @@
 <!--[if lt IE 10]>
 <div class="browserupgrade">
   <span class="fa fa-frown-o"></span>
-  <p>Sorry, your browser is out of date and not supported by MyUW. Learn how to upgrade <a href="http://browsehappy.com/">here</a>.</p>
+  <p>Sorry, your browser is out of date and not supported by MyUW.<br/>Learn how to upgrade <a href="http://browsehappy.com/">here</a>.</p>
 </div>
 <![endif]-->
 <noscript>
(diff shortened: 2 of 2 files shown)


timlevett:
Seems legit, :+1:

vertein:
Does seem legit.  I'm torn between wisc specific page and general page...  :+1:

keirserrie:
I like this solution, but I'd like to understand browser support at UWM a bit better. Discuss tomorrow after standup?

timlevett:
@keirserrie  - This is the kb article for MyUW : https://kb.wisc.edu/page.php?id=12737

apetro:
Browse Happy is, well, happy-making, but *I think we should have it link to UW documentation.*  If the right advice having explained the issue in that documentation is to visit something like Browse Happy to upgrade, well, then that documentation can link to Browse Happy, and in any case we can tweak documentation without cutting a release.

The issue is maybe a little more nuanced in that the user may not have privileges to upgrade the web browser on the terminal they are using, or the browser version may be frozen by their needing to use the old browser to access something else.  Presumably [UW documentation](https://kb.wisc.edu/page.php?id=12737) can try to address those nuances -- 

> Before upgrading your browser, we recommend that you contact your local IT support staff for any application incompatibilities that may be important to you.

That linked KB article is System flavored, though, and the context here is Madison, so maybe we need a new KB article for this purpose.

keirserrie:
Text tweaking here--

https://docs.google.com/a/wisc.edu/document/d/1LTEhRxx2UlsdDUP44rQlx21jb3eDTKUzRGy1MDZSDw8/edit?usp=sharing

jhanstra:
Updated:
image

This would show for IE 9 users too - it wouldn't show any MyUW functionality at all. Did we decide that's what we want to do? Once we have a KB doc link, I can swap that in. @apetro or @keirserrie, are you working on that or do you want me to create it? (If so, I'll need help setting up a page).

keirserrie:
@jhanstra we're going to test it on IE9 and see just how crappy it is, then make the call.

timlevett:
:+1:

keirserrie:
Thoughts on new wording here? Less judgy on the "out of date" stuff, prompts first to upgrade, then switch back.

:frowning:
Sorry, MyUW no longer supports your browser.

Learn how to upgrade your browser here.
Can’t upgrade? Switch back to the old version of MyUW here.

vertein:
OLD version of MyUW???  Now who's being judgy, @keirserrie?

apetro:
Eh.  

* I think I'm uncomfortable calling classic "the old version" until it is no longer default.  
* I'd also nuance it as "MyUW beta does not support your browser", not just "MyUW does not support your browser", because MyUW classic is MyUW, until it isn't. :stuck_out_tongue_winking_eye: 
* and I'm on a bender about the "here" links

So I'd suggest something more like

:frowning:
Sorry, MyUW beta does not support your browser.

[Learn how to upgrade your browser]().
Can’t upgrade? [Switch back to the old version of MyUW]().

but I'm not interested in insisting on any of these details; continuing to trust your professional UX judgment. :necktie:

keirserrie:
Andrew, I like your changes. Tweaked to avoid "old."

:frowning:
Sorry, MyUW beta does not support your browser.

[Learn how to upgrade your browser.]()
Can’t upgrade? [Switch back to MyUW classic.]()

timlevett:
:+1:

apetro:
Here's the URL to the newly minted KB article:

https://kb.wisc.edu/myuw/page.php?id=51345

I'll give the content there another love pass merging in @keirserrie 's fine ideas etc., but that should be enough to define the URL.

jhanstra:
Merging!

PR #221, opened 2015-05-07
MUMUP-1853 : Exclusive max mode

timlevett:
I was trying to think of more fun quick things we could do for the Apereo seminar, and I got to thinking about rendering portlets in exclusive mode, kind of like the static page we have on `/web`. One thing led to another and... now there is a PR. 

#### Woots
+ Behind beta flag
+ Works with uPortal and MyUW
+ Flippin' works
+ JavaScript works (thanks to trusty `$sce`)

#### Wohs
+ Forms, buttons, links that goto /portal (or /uPortal) render in exclusive mode on click, so... we have to deal with that sometime
+ Some portlets look funky
+ Didn't touch widgets, so beta flag is only for list view
+ Didn't setup add to favorite function (not hard tho)
+ Didn't setup rating (not hard tho)

#### Pictures, gifs, omg

http://goo.gl/yd3buZ

http://goo.gl/2qlwK2

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
+++ b/angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
@@ -16,7 +16,8 @@ define(['angular'], function(angular) {
             exampleWidgets: false,
             layoutMode : 'list', //other option is 'widgets'
             gravatarEmail : null,
-            useGravatar : false
+            useGravatar : false,
+            webPortletRender : false
             };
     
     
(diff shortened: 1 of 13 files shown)


apetro:
:+1: .  

It's behind a beta flag, so it isn't going to hurt anything, and even with the limitations, it's an interesting demo.

Most importantly, and the reason I'd favor merge sooner rather than later, this sets up some conversations with Unicon et al. wrt the Exclusive-mode portlet rendering they're looking at.

We can circle back on the fit and finish on this if/when it becomes important to ship this feature for reals in MyUW.  For now, more than enough here to have great Apereo conference collaborations upon.

timlevett:
I agree @apetro, I'll create a Jira and slip it into the sprint.

timlevett:
Created followup Jira for the wohs: https://jira.doit.wisc.edu/jira/browse/MUMUP-1854

PR #223, opened 2015-05-29
Update ui-bootstrap

paulerickson:
When I initially made the requirejs config, I missed uw-ui-toolkit.js and bootstrap.js, thinking these projects were CSS only.

This change makes it convenient to include uw-ui-toolkit _(and its transitive dependency on jQuery)_ by name; it does **not** automatically include it everywhere.

I defined uw-ui-toolkit, but **not** bootstrap, because uw-ui-toolkit appears to have been copied from bootstrap (v3.1.1).  Should bootstrap be included?  Is uw-ui-toolkit meant to coexist with it or replace it?


timlevett:
I think the `uw-ui-toolkit.js` is just bootstrap. We already have a bootstrap js with the `ui-bootstrap`. I'm not sure the two will play nice together.

paulerickson:
Good point, @timlevett .  I just googled and found one conflict between the two that was actually patched last year (point is - wasn't closed as "won't fix"), so I think they should be able to coexist, but certainly with some extra risk.

`uw-ui-toolkit.js` has about half as many lines as `bootstrap.js` 3.1.1; maybe it has only had some functions removed?  There are tons of other differences, but it might just be that `uw-ui-toolkit.js` was reformatted.  I don't spot any uw-specific looking code.

*anyway*, should `ui-bootstrap.js` be preferred to either bootstrap or uw-ui-toolkit?  If so, I can update this PR with a comment for uw-ui-toolkit, and maybe remove bootstrap altogether.

paulerickson:
For what it's worth, dnumber includes both `bootstrap.min.js` and `ui-bootstrap.min.js`, but not `uw-ui-toolkit.js`

timlevett:
I believe `ui-bootstrap.js` would be preferred as it was written as a solution for angularjs apps with bootstrap. 

https://angular-ui.github.io/bootstrap/

paulerickson:
Is there custom code or another reason why `ui-bootstrap.js` is included directly instead of through bower?  I assumed *"no"* and put it in bower with the latest version.

I removed the path to `uw-ui-toolkit.js` with a comment so the next person (me, once I forget) will know not to use it.

timlevett:
Nope, no reason not using bower. I think when we added it we were not using bower yet. #alwaysImproving

:+1:

PR #225, opened 2015-06-04
Use correct artifact for ui-bootstrap

paulerickson:
`ui-bootstrap.js` references a bunch of html templates, but does not define them.  So, for instance, you'd get a 404 on "templates/accordion/accordion.html".  The cryptically named `ui-bootstrap-tpls.js` is the one to use, because it includes the templates.

I also noticed the file had CRLF line endings (my bad) so I replaced them with LF.


timlevett:
:+1:

PR #226, opened 2015-06-08
Add in root index file for jetty run

timlevett:
+ Adds in an `index.html` to the root context

image


vertein:
After you fix the title, feel free to merge. :+1:

README.md
```diff
@@ -25,7 +25,7 @@ This is the MyUW home page. It uses the frame as a base then adds in the layout,
 To deploy the home build from the base directory described above. Then `cd ./angularjs-portal-home` and run `mvn tomcat7:redeploy` (assuming you have auto deploy configured).  The home will now be deployed to `/web`.
 
 ### Running
-To run Frame or Home in isolation with the jetty-maven plugin, run `mvn jetty:run` or `mvn jetty:run-war` from the submodule directory.
+To run simply type 'mvn jetty:run` from the root directory. By default jetty runs on port 8080.
```
paulerickson: I think that you can't actually use `mvn jetty:run` from the parent dir unless you first run `mvn package` (or something else that encompasses that lifecycle stage), although I haven't verified.  `mvn jetty:run` serves from the `src` directory, but doesn't have a way of recursing into the subdirs to do the same (as far as I know).

src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,17 @@
...
+
+<web-app 
+    xmlns="http://java.sun.com/xml/ns/javaee"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
+    version="2.5">
+  <display-name>ROOT Mock Redirector</display-name>
+  
+  <welcome-file-list>
+    <welcome-file>index.html</welcome-file>
```
paulerickson: This is one of the defaults, I think.  I wonder if a web.xml is actually needed for this?

PR #227, opened 2015-06-08
Css spring cleaning

timlevett:
+ Removed unused LESS files
+ Removed unused font-awesome css file (we are using bower now)
+ Cleaned up usage of hex vs color variables defined in variables.less

No visual changes


paulerickson:
:+1: If less is more, then less less is... even better!

PR #228, opened 2015-06-09
Set up karma testing

paulerickson:
I ran `bower update`, forgetting that the test dependencies are actually in package.json, so the changeset looks bigger than it really is.  I figure it's good to keep updated, but let me know if you'd rather I reverted that part.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/main.js
+++ b/angularjs-portal-frame/src/main/webapp/main.js
@@ -1,38 +1,11 @@
-require.config({
+'use strict';
+require(['./config'], function(config) {
 
-    packages: [
-        'portal',
-        'my-app'
-    ],
+    require.config(config);
 
-    paths: {
-        'angular'       : "bower_components/angular/angular",
-        'app-config'    : "js/app-config",
-        'jquery'        : "bower_components/jquery/dist/jquery.min",
-        'jquery-ui'     : "bower_components/jquery-ui/jquery-ui.min",
-        'ngRoute'       : "bower_components/angular-route/angular-route.min",
-        'ngSanitize'    : "bower_components/angular-sanitize/angular-sanitize.min",
-        'ngStorage'     : "bower_components/ngstorage/ngStorage.min",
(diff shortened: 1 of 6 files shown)


timlevett:
Hmmm, i guess which files should we look at? hard to read through the weeds

timlevett:
or show how to run locally so we can pull down and checkout?

apetro:
I like dependency updates.  I like the tests working.  I'd like these better in separate commits.

paulerickson:
Yeah, it's pretty gross... I broke into two commits, and I'll move the bower update to its own branch/PR.

paulerickson:
Sorry about that, should be readable now

paulerickson:
@timlevett, with `/web` overlaying this project, all you should have to do is define karma tests anywhere in the project, with requirejs `define` headers and filenames ending in 'test' or 'spec' (like `service_test.js` or `widgetSpec.js`, etc.).

paulerickson:
I'm running the tests by running `npm test` from `src/main/webapp`, btw.

angularjs-portal-frame/src/main/webapp/main.js
```diff
@@ -1,38 +1,11 @@
-require.config({
+'use strict';
+require(['./config'], function(config) {
 
-    packages: [
-        'portal',
-        'my-app'
-    ],
+    require.config(config);
```
paulerickson: I broke this into a separate file so that the main.js for tests can use it too, and keep as much code as possible in common.

timlevett:
Looks good to me. Does breaking out the config.js require any frame app changes or should it just work?

paulerickson:
It's still the first thing to be evaluated, before bootstrapping the application into angular, so it works just the same as before.

It should be completely invisible to downstream apps like angularjs-portal-home, since they overlay the `my-app` package a level below this, not these top-level files themselves.

timlevett:
Sounds awesome, nice work @paulerickson . :+1:

PR #230, opened 2015-06-11
Fix search bar pixel

jhanstra:
This doesn't completely fix the issue - it's still off by a pixel at other zooms, but it fixes it for 100% zoom, which is what most people will be using.

It's a bug with Bootstrap, which is also a tad wonky at different zooms.

MyUW, 100% Before:
image

MyUW, 100% After:
image

MyUW, 110% After:
image

Bootstrap, 110%:
image


timlevett:
:+1: for 100% zoom fix. Most people will be at 100% like you said.

apetro:
:+1:

PR #233, opened 2015-06-11
MUMUP-1876 : Fix style so that you can have longer than normal app names

timlevett:
### Before 
http://goo.gl/e6oznh
### After

http://goo.gl/5zlaIg

http://goo.gl/jyzVyk

_Please note that this doesn't actually change the name, just lets frame apps have the ability to be more wordy in there title._


vertein:
:+1:

apetro:
:+1:

PR #235, opened 2015-06-16
MUMUP-1911 : Switched over to new proxy service

timlevett:
#### Motivation
A few other groups needed this external service, so I did that here: https://github.com/UW-Madison-DoIT/json-proxy-service, and I needed to test to make sure it worked, so i did that here. :)

#### What Changed?
This PR "eats our own dog food", but the end result is nothing changed.

#### Reason for pending merge
I would like to get a real release of the proxy service in maven central before pushing this so it doesn't break the build as it currently isn't deployed anywhere.


apetro:
:+1: . Agree with preference for depending on a real release rather than a `SNAPSHOT`.

I love the eating own :dog2: :pizza: .  I also love the factoring out much smaller cohesive pieces of the solution as their own versioned artifacts in their own repos.  This is an incredibly healthy, necessary, important refactoring, one that enables re-use, makes tooling more effective, and makes it easier to understand what software is using this utility how.  Hurray for refactoring and its enabling agility.

PR #237, opened 2015-07-01
Replace ngRoute with ui-router

paulerickson:
I've been reading for a while that [ui-router](https://github.com/angular-ui/ui-router) is better than ngRoute in every way, and I finally had a use case to try it out in bucky-backup.

This changeset is to swap out the router with functionality intact, rather than leverage new features or restructure things.

Some benefits of ui-router that we want to incorporate (in subsequent PRs):
* **Nestable views**
  * Minimize duplication in templates and only reload the part of the page that you need (which might be a tile within a section within the main page, e.g.)
  * Nested URLs can be built dynamically - so child views don't need to be tightly coupled with the URLs of parents (yay, modularity!)
* **Named / parallel views**
  * Page could be split into multiple views, like sidebar, header, page, etc., and an app could potentially attach to a named view within the sidebar, for instance.
* **Passing state between views**
  * e.g. for wizard-style workflows


smargovsky:
Just my 2 cents, if we are going to replace such an important piece as the router, why not look forward and use the new router ( 1.4 and 2 ) : https://angular.github.io/router/

paulerickson:
I am excited for "new router", but unfortunately it didn't make the 1.4 release as planned and is still under development for the 1.5 release (correct me if I'm mistaken!)
See https://github.com/angular/router

From what I read, the syntax has changed in the past month or so making half the articles out there inaccurate, so I feel nervous about moving to it until it's more established anyway.

timlevett:
I would prefer to stay as vanilla as possible in frame. I too am excited for the real angular router ngRoute 2.x which should be out really soon. I'm not down voting this, just voicing my opinion.

apetro:
I too would prefer to avoid a ui-router side quest in favor of picking up ngRoute v2 when baked.  Complexity has a cost.

@paulerickson I'd be very interested in understanding what Bucky Backup needs are blocked on this, if any, and what the next best alternative is.  I could see myself being talked into the ui-router side quest if it enables things that are essential in frame apps sooner than the more boring wait-for-ngRoute-v2 path will allow.

paulerickson:
@apetro what spurred me was wanting to explore breaking up long forms that conditionally show/hide sections and using a wizard-style progression instead.  Nesting small form components/states within the overarching form seems like a clean way to go about it, but there are certainly alternatives... I could use one bigger controller and many directives, or separate controllers but use `$rootscope`, or the `angular-wizard` plugin, etc.

I am okay with passing on this and waiting for the brand-name router, _but_, just to play devil's advocate - where is the complexity here; wouldn't ui-router be an ideal transitional phase?
It seems to me that the syntactical changes are very minor, and the real work is structural.  Since componetizing was done with the requirejs PR, this changeset is just syntax; if we restructure templates to leverage nesting and naming, then integrating new `$router` will likewise have a smaller scope of replacing names here and there.

timlevett:
A lot of these changes would now need to happen in uw-frame. I'm going to close this PR in light of that.

PR #241, opened 2015-07-02
MUMUP-1930 update README re providing endpoint.properties .

apetro:
If you don't provide an `endpoint.properties`, the `/web` war file will not Spring applicationContext init, and so fails to come up under e.g. `jetty:run`.

This PR updates `README.md` to document that one needs to do this before building.

Went for documentation of the step rather than just going ahead and providing an empty `endpoint.properties`, since a `.gitignore` had carefully excluded `endpoint.properties`.


vertein:
:+1: Way favor this over an empty endpoint.properties. :)

PR #243, opened 2015-07-08
MUMUP-1943 : Add configuration to get frame apps search to work

timlevett:
Basically adds a redirect if its a frame app.  Has a shortcut for `/web` to avoid page refresh.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/js/app-config.js
+++ b/angularjs-portal-frame/src/main/webapp/js/app-config.js
@@ -17,6 +17,10 @@ define(['angular'], function(angular) {
             'crest' : 'img/uwcrest_web_sm.png',
             'crestalt' : 'UW Crest'
         })
+        .constant('SEARCH',{
+            'isWeb' : false,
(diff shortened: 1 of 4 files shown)


angularjs-portal-frame/src/main/webapp/js/app-config.js
```diff
@@ -17,6 +17,10 @@ define(['angular'], function(angular) {
             'crest' : 'img/uwcrest_web_sm.png',
             'crestalt' : 'UW Crest'
         })
+        .constant('SEARCH',{
+            'isWeb' : false,
```
apetro: I wonder if web-ness is going to end up being a concept applicable more broadly than in search.  I'm fine with it starting here where angularjs-portal is prepared to take advantage of the concept today, and refactor it out if it ever turns out to be a wider thing.

apetro:
:+1:

angularjs-portal-frame/src/main/webapp/js/app-config.js
timlevett: Logical, I'll leave as is until that day come.

PR #250, opened 2015-07-15
MUMUP-1856 : Notifications by group

timlevett:
+ Refactored the controller to use `$scope`
+ Externalized some config for notifications for frame apps
+ Added `getNotificationsByGroups` to the `notificationsService`. This call does 2 http requests. 1 for groups, the other for notifications. Then it postprocesses to filter notifications by groups.
+ Added a new dummy notification for student group for testing

_No visual changes_

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/js/app-config.js
+++ b/angularjs-portal-frame/src/main/webapp/js/app-config.js
@@ -23,7 +23,10 @@ define(['angular'], function(angular) {
             'isWeb' : false,
             'searchURL' : '/web/apps/search/'
         })
-        ;
+        .constant('NOTIFICATION', {
+            'groupFiltering' : true,
+            'notificationFullURL' : 'notifications'
+        });
 
     return config;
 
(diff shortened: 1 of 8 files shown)


vertein:
so clean, much wow :+1:

angularjs-portal-frame/src/main/webapp/js/app-config.js
```diff
@@ -22,9 +23,11 @@ define(['angular'], function(angular) {
             'isWeb' : false,
             'searchURL' : '/web/apps/search/'
         })
-        ;
+        .constant('NOTIFICATION', {
+            'groupFiltering' : true,
```
apetro: Interesting that whether it filters by groups is configurable.  Would one ever not filter by groups?  In a future when the notifications URL does filtering server-side so that there's no further filtering to do client-side?

angularjs-portal-frame/src/main/webapp/portal/notifications/controllers.js
```diff
@@ -4,44 +4,53 @@ define(['angular'], function(angular) {
...
+      $scope.count--;
     }
+
+    var init = function(){
+      $scope.notifications = [];
+      $scope.count = 0;
+      $scope.isEmpty = false;
+      $scope.notificationUrl = NOTIFICATION.notificationFullURL;
+
+      if(NOTIFICATION.groupFiltering) {
```
apetro: Might be good to default to filtering by group if `NOTIFICATION.groupFiltering` is not set, so as to fail safe on ambiguous configuration.

angularjs-portal-frame/src/main/webapp/js/app-config.js
timlevett: Was thinking more for re-usability for frame applications

PR #251, opened 2015-07-15
MUMUP-1955 : Add in SSN missing notification for that group

timlevett:
image


timlevett:
I added this also to the overlay, so we can finalize the notification wording at a later date.

vertein:
:+1:

PR #252, opened 2015-07-15
MUMUP-1912 : Add in back to classic link and config

timlevett:
+ Adds config for `feedback`
+ Adds config for `back to classic`
+ Adds config for `what's new`
+ Adds back to classic to the footer, if its configured

image
+ Hides if not configured or `null`

image

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/css/buckyless/footer.less
+++ b/angularjs-portal-frame/src/main/webapp/css/buckyless/footer.less
@@ -12,10 +12,18 @@ footer .portal-power {
 }
 
 footer a {
-	color: @footer-text-color;	
+	color: @footer-text-color;
 	text-decoration: none;
 }
 
+footer .underline a {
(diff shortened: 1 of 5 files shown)


jhanstra:
Thoughts on changing wording to 'Switch to Classic MyUW'? Other than that, :+1:. :D

keirserrie:
I had it just as "MyUW Classic" but am flexible. https://docs.google.com/document/d/14X1v0N0TPWecEkVbLPwAw3YDZv0LMZ3SnQEwMwXrorg/edit

keirserrie:
I want to make it so it's easy, but not obvious for someone who isn't looking for it. So you don't have somebody in the footer say "hey, i can switch back? sweet!" and we lose another user.

jhanstra:
@keirserrie I like "MyUW Classic", agree with your logic

timlevett:
neat, fixing

PR #255, opened 2015-07-16
MUMUP-1948 : Remove beta header, will fully delete at a later date

timlevett:
+ Removes `beta-header` from the frame
image
+ Externalized the sub title (beta) so people can choose what they want
image

I just removed the directive and less import, but didn't delete the source. Just in case we wanted to utilize these things in the future.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/js/app-config.js
+++ b/angularjs-portal-home/src/main/webapp/js/app-config.js
@@ -26,7 +26,8 @@ define(['angular'], function(angular) {
         .constant('NAMES', {
             'title' : 'MyUW',
             'crest' : 'img/uwcrest_web_sm.png',
-            'crestalt' : 'UW Crest'
+            'crestalt' : 'UW Crest',
+            'sublogo' : ''
         })
         .constant('SEARCH',{
             'isWeb' : true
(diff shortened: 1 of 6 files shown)


keirserrie:
:+1:  I know this is scope creep, but has there ever been talk of more of a logo there instead of just text?

timlevett:
@keirserrie - there hasn't, but I'm sure people might be interested.

angularjs-portal-home/src/main/webapp/js/app-config.js
```diff
@@ -26,7 +26,8 @@ define(['angular'], function(angular) {
         .constant('NAMES', {
             'title' : 'MyUW',
             'crest' : 'img/uwcrest_web_sm.png',
-            'crestalt' : 'UW Crest'
+            'crestalt' : 'UW Crest',
+            'sublogo' : ''
```
apetro: Missed this subtlety previously, but may want to keep the "beta" sub-logo.  Cf. [design recommendations](https://docs.google.com/a/wisc.edu/document/d/14X1v0N0TPWecEkVbLPwAw3YDZv0LMZ3SnQEwMwXrorg/edit?usp=sharing), 

> The “beta” text on the logo will remain as an acknowledgement that this is a work in progress, buying some leniency from both users and content providers:

> When MyUW Classic is retired, the beta text will be removed from the MyUW logo.

timlevett: app-config.js is in the overlay. Enjoy.

apetro: My bad; design had changed and I'd forgotten the conversation.  Dropping the Beta sub-logo works.

PR #256, opened 2015-07-16
My uw apps page

jiayinjx:
MyUW Apps page: add description to all "Launch" button to clear that which app will be launched after press that button.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -13,7 +13,7 @@
     </div>
     
     <div class="widget-title">
-      <h4>{{ ::portlet.title }}</h4>
+      <h4 id="appTitle">{{ ::portlet.title }}</h4>
     </div>
   </div>
   
@@ -45,11 +45,11 @@
   </div>
   
   <!-- For basic apps (not a widget, not simple content, no pithy content), show only an icon -->
-  <a ng-if="'NORMAL' === widgetCtrl.portletType(portlet)" href="{{::portlet.url}}" target="{{::portlet.target}}">
+  <a aria-labelledby="appTitle" ng-if="'NORMAL' === widgetCtrl.portletType(portlet)" href="{{::portlet.url}}" target="{{::portlet.target}}">
     <div class="widget-icon-container">
       <portlet-icon></portlet-icon>
     </div>
-    <button class="btn btn-default launch-app-button">Launch full app</button>
+    <button aria-labelledby="appTitle" class="btn btn-default launch-app-button">Launch full app</button>
   </a>
   
   <!-- For pithy content, display the pithy content -->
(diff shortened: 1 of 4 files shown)


keirserrie:
We need the aria text to be more descriptive of what's going to happen here, so instead of the screen reader saying "weather," it should say "Go to weather." For simplicity's sake, I recommend that we also change the button text to say the same thing, so it's always: Go to [app name]

I'm recommending "go to" instead of "launch" as a more generic and easily understood term.

Anyone have thoughts on this?

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-portlet.html
```diff
@@ -1,5 +1,5 @@
 <div class="portlet-list-item" >
-  <h2>
+  <h2 id="appTitle">
```
timlevett: The id of a field must be unique across the whole document (the whole page). Read http://www.w3.org/TR/html4/struct/global.html#adef-id for more details.

I would suggest using a programmatic id, like `appTitle{{::portlet.title}}` or something like that.

vertein:
@keirserrie IRT your comment.  I've read that launch is kind of a confusing word.

jiayinjx:
@keirserrie, @timlevett , @vertein, I can add another text without change the content in the button, and let the screen reader reads " go to [appTitle{{::portlet.title}}]. Or if you suggest to change the text inside the button?

keirserrie:
I would prefer also changing the button text, but we can discuss after standup.

jiayinjx:
1. make the apps' title able to be pointed out by TAB
2. make the add to favorites button reads "Add to favorites appTitle{{::portlet.title}}"
    similarly, already in favorites button reads" Already in favorites appTitle{{::portlet.title}}"
3. launch button: "go to appTitle{{::portlet.title}}"

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -9,11 +9,11 @@
       tooltip-popup-delay="200"></i>
     </div>
     <div class='widget-remove'>
-      <i title="Remove" class="fa fa-times portlet-options" ng-click="widgetCtrl.removePortlet(portlet.nodeId, portlet.title)"></i>
+      <i title="Remove" class="fa fa-times portlet-options" ng-click="widgetCtrl.removePortlet(portlet.nodeId, portlet.title)"><a aria-label="Remove this app" href="#"></a></i>
     </div>
     
     <div class="widget-title">
-      <h4>{{ ::portlet.title }}</h4>
+      <h4 id="appTitle">{{ ::portlet.title }}</h4>
```
timlevett: You can only have one id that is titled "appTitle". This will break the page as this widget-card is used multiple times.

jiayinjx: Already update it to appTitle{{::portlet.title}} in widge-card.html and make this available in default-card.html which is for icon view, too.

timlevett: I'm not sure I follow.

jiayinjx: I improved it in another branch, might be the reason it doesn't shows up here.....

timlevett: Hmmm, should we close this PR in favor of the later pull request. We should really only have one PR for each set of changes.

jiayinjx: of course, I changed the branch due to I found it was not in the correct branch. Sorry for miss this up.

timlevett:
Closing as a duplicate.

PR #257, opened 2015-07-20
Cleanup Unused code

timlevett:
+ Removes everything under `bucky` in favor of `buckyless`
+ Removes extra bootstrap less
+ Removes old config files we no longer use
+ Fixes the `master_app-config.js`
+ Removes test results because that is just weird


jhanstra:
:+1: :+1: :+1: :+1: :+1:

timlevett:
much thumb, wow. :dog2:

vertein:
no kidding, much thumb

vertein:
I was 80% of the way through, so I thought I might as well finish.  Here's another :+1:

PR #258, opened 2015-07-20
MUMUP-1925 : Add in opt out link on top of document for sr

timlevett:
+ Fixed some spacing that was weird
+ Adds in a div for `sr-only` (bootstrappy thing) based on if they have a `back2classicURL` set. (good for frame apps opt out)


vertein:
:+1:

PR #259, opened 2015-07-20
Home page icons view

jiayinjx:
1. make the app can be delete from the home-page icon view by the cross symbol
2. add description to each app which will let the user know which app they are at before launching.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -13,7 +13,7 @@
     </div>
     
     <div class="widget-title">
-      <h4>{{ ::portlet.title }}</h4>
+      <h4 id="appTitle">{{ ::portlet.title }}</h4>
     </div>
   </div>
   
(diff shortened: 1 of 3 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -9,14 +9,14 @@
...
+      <i title="Remove" class="fa fa-times portlet-options" ng-click="widgetCtrl.removePortlet(portlet.nodeId, portlet.title)"><a aria-label="Remove this app" href="#"></a></i>
     </div>
     
     <div class="widget-title">
-      <h4>{{ ::portlet.title }}</h4>
+      <h4 id="appTitle{{::portlet.title}}" aria-labelledby="appTitle{{::portlet.title}}" tabindex="0">{{ ::portlet.title }}</h4>
     </div>
   </div>
-  
+  <sub hidden="" id="goToApps">go to</sub>
```
timlevett: Again, the id must be unique across the whole document. The `widget-card` partial is used many times.  Also if you want to hide for everything but a screen reader just use `class='sr-only'`. Its a bootstrap class. Reason for the `<sub>` tag?

jiayinjx: I didn't see any id like goToApps...... so, change to id="goToApp{{::portlet.title}" work?

timlevett: I see what you mean, so you are concatenating the fields together to give a better aria-label in the end. I would still suggest the `id="goToApp{{::portlet.title}"`.

jiayinjx: Yes, 
widgecard
so is that ok if I just change the id and leave <sub> tag there like this?

timlevett: Yes, that sounds good

PR #260, opened 2015-07-20
Fix hover state of notification on full

timlevett:
We had this hover move thing going on as part of the demo. Just cleaning it up a bit so just the background color shade changes a bit.

### Before
http://goo.gl/8cZucT

### After
http://goo.gl/Aadmtd


apetro:
:+1:

vertein:
:shipit:

PR #261, opened 2015-07-20
MUMUP-1947: Welcome to beta

jhanstra:
image

- User will only see the popup one time after this code is released to production. After it is shown and the user clicks out of it, a local variable will record that it has been seen and will not show again. Do we want to add in something that checks the date to see if it is August 11 yet? So that we can push a release sooner?

- Popup does not show up in frame, only in /web

- Still needs link to KnowledgeBase article

- Popup does not show on mobile


angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
```diff
@@ -61,6 +62,19 @@ define(['angular'], function(angular) {
       that.user = result;
     });
   }]);
+  
+  app.controller('WelcomeController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', '$modal', function($localStorage, $sessionStorage, $scope, $document, NAMES, $modal) {
+    $scope.openModal = function() {
+      if ($localStorage.hasSeenWelcome === false) {
+        $modal.open({
+          animation: $scope.animationsEnabled,
+          templateUrl: require.toUrl('../../my-app/layout/partials/welcome.html'),
```
timlevett: Hmmm, this breaks frame apps if the child app didn't have this file. Why not have the partial in frame?

timlevett: Confirmed, if you hit /frame and don't have the flag set, it throws:

```
Error: [$compile:tpload] Failed to load template: ./my-app/layout/partials/welcome.html (HTTP status: 404 Not Found)
http://errors.angularjs.org/1.4.0/$compile/tpload?p0=.%2Fmy-app%2Flayout%2Fpartials%2Fwelcome.html&p1=404&p2=Not%20Found
```

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
```diff
@@ -61,6 +62,19 @@ define(['angular'], function(angular) {
       that.user = result;
     });
   }]);
+  
+  app.controller('WelcomeController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', '$modal', function($localStorage, $sessionStorage, $scope, $document, NAMES, $modal) {
+    $scope.openModal = function() {
+      if ($localStorage.hasSeenWelcome === false) {
```
timlevett: You could just set this to `if ($localStorage.hasSeenWelcome)`. #falsyforlife

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
jhanstra: Isn't that just in the console though? I don't think anything user-facing breaks, right? I originally had it in frame, but then wouldn't every app using frame (student-parent-financial-portal, dnumber, etc) show the welcome popup for MyUW? I think that might be weird.

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
jhanstra: Good call #lesscharacterslessproblems

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
-define(['angular'], function(angular) {
+define(['angular','require'], function(angular, require) {
   var app = angular.module('portal.main.controllers', []);
 
-  app.controller('MainController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', 'MISC_URLS', function($localStorage, $sessionStorage, $scope, $document, NAMES, MISC_URLS) {
+  app.controller('MainController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', 'MISC_URLS', '$modal', function($localStorage, $sessionStorage, $scope, $document, NAMES, MISC_URLS, $modal) {
```
timlevett: Minor, but is $modal used in the main controller?

jhanstra: Nope, that was from my first try around. Good catch.

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
```diff
@@ -61,6 +62,19 @@ define(['angular'], function(angular) {
       that.user = result;
     });
   }]);
+  
+  app.controller('WelcomeController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', '$modal', function($localStorage, $sessionStorage, $scope, $document, NAMES, $modal) {
+    $scope.openModal = function() {
+      if (!$localStorage.hasSeenWelcome) {
+        $modal.open({
+          animation: $scope.animationsEnabled,
+          templateUrl: require.toUrl('../../my-app/layout/partials/welcome.html'),
```
timlevett: Weird, my comment went away :(

angularjs-portal-home/src/main/webapp/my-app/layout/partials/welcome.html
```diff
@@ -0,0 +1,8 @@
+<div class="welcome-popup center">
+  <h1>Welcome to the new MyUW. Here are the basics.</h1>
+  <img src="img/myuw-welcome.png">
```
apetro: What can be done to provide equivalent access to non-sighted users in getting the basics about the new MyUW?  Description tagging this image?  Alternative text description of the basics of MyUW available somewhere else, perhaps  linked from a screen-reader-only link at the top of MyUW?

I'd like MyUW to do something to achieve "equivalent access" accessibility of the welcome content, and I'd like to be able to articulate what that something is when describing how MyUW is attending to accessibility.

timlevett: Is this in scope? sounds like a making of a great story.

apetro: Equivalent-access accessibility was [discussed when this story was groomed](https://jira.doit.wisc.edu/jira/browse/MUMUP-1947?focusedCommentId=597978&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-597978), as I recall.

timlevett: :+1:

jhanstra: :+1: also. What do we think about this below the image? I like @timlevett's idea of an sr-only div. @keirserrie and @jiayinjx, thoughts on wording and whether this is helpful for screen reader users?

```
<div class="sr-only">
    <p>The new MyUW makes it easier to access your content. You can switch between simple icons and more informative widgets using 'Icon View' and 'Widget View' toggles. Each of the icons on the home page links to your MyUW content, and widgets can display a dashboard of information on the home page. If you are having trouble finding what you're looking for, you can use the top search box to search through all of MyUW's content, and save it to your home page from there using the 'Add to Favorites' button.</p>
</div>
```

jiayinjx: The content will not be read by screen reader.
add tabindex="0" id="readWelcomeContent" aria-labelledby="readWelcomeContent" inside the <div> tag will make this work - make the content can be focused by TAB in navigation order and then read out when the focus on it.
And make an unique id name (the id name above is just an example)

timlevett:
Tested out on mobile, the modal still comes up even with the div hidden. So we just get a grayed out portal with no escape

http://goo.gl/BhbLQo

Test case: Enable mobile emulator, goto `/web/settings`, clear local storage, refresh page

angularjs-portal-frame/src/main/webapp/portal/main/partials/welcome.html
```diff
@@ -0,0 +1,11 @@
+<div class="welcome-popup center">
+  <h1>Welcome to the new MyUW. Here are the basics.</h1>
+  <img src="img/myuw-welcome.png">
+	<div class="sr-only">
+		<p>The new MyUW makes it easier to access your content. You can switch between simple icons and more informative widgets using 'Icon View' and 'Widget View' toggles. Each of the icons on the home page links to your MyUW content, and widgets can display a dashboard of information on the home page. If you are having trouble finding what you're looking for, you can use the top search box to search through all of MyUW's content, and save it to your home page from there using the 'Add to Favorites' button.</p>
+	</div>
+  <div>
+    <button class="btn btn-lg btn-primary" ng-click="$dismiss()">Continue to MyUW</button>
+  </div>
+  <p>Or learn more at the <a href="#">KnowledgeBase</a></p>
```
apetro: Missing link acknowledged in the pull request description.  Nonetheless, flagging with this comment : Learn more link needs to either disappear or link somewhere before this is merged so that the code stays shippable. :smile:

angularjs-portal-frame/src/main/webapp/portal/main/controllers.js
```diff
@@ -61,6 +62,19 @@ define(['angular'], function(angular) {
...
+  
+  app.controller('WelcomeController', ['$localStorage', '$sessionStorage','$scope', '$document', 'APP_FLAGS', '$modal', function($localStorage, $sessionStorage, $scope, $document, APP_FLAGS, $modal) {
+    $scope.openModal = function() {
+      if (APP_FLAGS.welcome && !$localStorage.hasSeenWelcome) {
+        $modal.open({
+          animation: $scope.animationsEnabled,
+          templateUrl: require.toUrl('./partials/welcome.html'),
+          size: 'lg',
+        });
+        $localStorage.hasSeenWelcome = true;
```
apetro: Probably an unimportant nitpick, but: would it be better to flip this bit on `$dismiss()` rather than on render, so that a user who hasn't dismissed this message sees it again if they reload the page, until they dismiss it?

jhanstra:
@timlevett it looks like you can still just click anywhere to get rid of it but that's still gross. I'll maybe make a lighter weight modal for mobile.

jhanstra:
Possible mobile experience, thoughts? I think the image will be too small to display on mobile. @keirserrie 

image

keirserrie:
@jhanstra this is good, just adjust line breaks to get rid of orphans and :+1:

timlevett:
@jhanstra - seems legit. Maybe just a link to the video/KB for mobile? The sentence seems redundant.

So my thought is:

image

jhanstra:
@timlevett, like it. 
image

angularjs-portal-home/src/main/webapp/my-app/layout/partials/welcome.html
jhanstra: Thanks @jiayinjx! Can you check my code again to see if I did it correctly?

jiayinjx: Had a look at it. Greet!

timlevett:
:+1:

vertein:
:+1:

angularjs-portal-frame/src/main/webapp/samples/welcome.json
```diff
@@ -0,0 +1,11 @@
...
+   "id" : 1,
+   "title" : "",
+   "img" : "",
+   "learnMoreURL" : "",
+	 "startYear" : 2015,
+	 "startMonth" : 7,
+	 "startDay" : 11,
+	 "accessibilityText" : "",
+	 "buttonText" : ""
+}
```
timlevett: Thoughts on just making this an array instead of an object? So `/frame` can just be `[]`.

Then in the controller you can just check the length, and if the length is > 0 then go through the date logic.

jhanstra: good idea

timlevett:
:+1:

angularjs-portal-frame/src/main/webapp/portal/main/partials/welcome-modal-template.html
```diff
@@ -0,0 +1,12 @@
+<div class="welcome-popup center">
+  <h1 ng-bind-html="welcome.title"></h1>
+  <img src="{{ welcome.img }}" class="hidden-xs">
+	<div class="sr-only" tabindex="0" id="welcomeContent" aria-labelledby="welcomeContent">
```
apetro: This `aria-labelledby="welcomeContent"` looks weird to me, in that the div marked as labelledby welcomeContent itself has the `id` welcomeContent, so isn't that saying it's labeled by itself?

Is there something that should be done to the `<img` tag itself to make it accessible?  Maybe that's the element that should have the `aria-labelledby` attribute on it?  And maybe it's more of an [`aria-describedby`](http://www.w3.org/WAI/GL/wiki/Using_aria-describedby_to_provide_descriptions_of_images), that is, more of a description than a label?

The technique here feels kind of like externalizing the text description of the image, what with it being designated `sr-only`.  As in, here's an image, and here's some text that describes that image that you might want to read alternatively to that image.  Potential in going further an externalizing it via the [`longdesc` technique](http://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/H45.html)?

angularjs-portal-frame/src/main/webapp/portal/main/partials/welcome-modal-template.html
```diff
@@ -0,0 +1,12 @@
...
+		<p>{{ welcome.accessibilityText }}</p>
+	</div>
+  <!-- <p class="visible-xs-* learn-more">Learn more at the <a href="#">KnowledgeBase.</a></p> -->
```
apetro: Idea: What if the display of the learn more links were predicated on the `learnMoreURL` data having a non-empty value? Such that this HTML template could ship ready to display the learn more link, and would display it if the `welcome.json` were to start including that value.  That might put this even closer to ready to go such that the KB URL could be added to just the one file if it emerges late towards the August 11th release.

jhanstra: That could be interesting. But the KB article would still only have to be added to one file, this one, currently. So it's probably more work to write that logic than to simply uncomment and stick in the links here.

apetro: That's fair.  This might be a refactor to pick up when going after the multi-page version of this, since it's going to become more plausible that some pages have learn more links and others don't.

angularjs-portal-frame/src/main/webapp/portal/main/partials/welcome-modal-template.html
jhanstra: I'm not sure how the accessibility for this works, I'll do some research and talk to @jiayinjx. The only reason I didn't put the description as alt text on the image is that it was a really long description, and it seemed too long for alt text. But I'm starting to think it would be better as alt text.

apetro: Cool.  I don't want to lose sight of the progress we're making: having this accessibility text at all is better than not having it, and is definitely progress, and is to be appreciated and commended. :cake: 

Looking around, I think this may be an opportunity to dig into the distinction between labels, which give non-visual users pithy handles to understand the objects on the page, and descriptions, which give non-visual users an alternative way to understand the content of the objects on the page.  Our very own [DoIT accessibility guidelines](http://www.doit.wisc.edu/outside-cms/accessibility/online-course/standards/images.htm) seem to have some relevant advice.

jiayinjx: if you want the screen reader just read the title or let them know it is a welcome pop-up window, you might want to add the id on the title(not the full text area) and use aria-labelledby; or if you want it reads all the text like a paragraph as long as the window pop-up, use a aria-describedby. My suggestion might be first let them know there is an welcome pop-up window, then they might use TAB to focus on the text. This might be helpful:http://www.w3.org/TR/wai-aria/states_and_properties

PR #262, opened 2015-07-21
MUMUP-1955 : remove alpha flag for notification. 

timlevett:
+ Removed beta notification flag in settings
+ Moved notification flag to app-config


apetro:
:+1:

PR #263, opened 2015-07-24
Better accessibility for welcome popup image

jhanstra:
Thoughts on this? `aria-describedby` seems like the best option to me. But using a long alt text or using the `longdesc` option also seem legitimate.


apetro:
:+1: I *think* this is a further improvement, just like having the description text at all was an improvement.  Will be fun to test this out on a screen reader or two.

PR #264, opened 2015-07-27
Remove gross transparent images

jhanstra:
Found some nasty 1px transparent images, squashed em.

I now realize that this was mock data, so it wasn't a big deal at all, but oh well - now it annoys me less :) #developerhappiness

Before:
image

After:
image


apetro:
:+1:  Hurray for clean, even in mock data.  Worth it.

PR #270, opened 2015-07-29
EC: Material

timlevett:
Food for thought:

http://goo.gl/x85kvW

Random thought, don't merge, just throwing this out there.

Using this: [ngMaterial](https://material.angularjs.org/latest/#/)


jhanstra:
Oh man. This is very interesting. I think using Material for the whole UI would make it look *much* different (and very googley), and probably not advisable so soon after people move to Beta. But we can definitely start taking advantage of some parts of it, see how we like it, and gradually move toward that if we want to. 

Where do other people see this being used in the future?

apetro:
(Placeholder for conversation about going beyond a style guide, beyond the current toolkit, to a whole user interface system, which is kind of what Google's Material Design is offering.)  

Short version: I think there's a lot of potential here; it's this kind of potential that's worrying me some that frame-based apps are being developed in some ways before we're ready for that; adopting [a system generally and Material Design specifically](https://www.youtube.com/watch?v=JvAEhQxr6rU) could go a long way to normalizing the experience across MyUW components in a way that is enabling and excellent rather than merely limiting.

I'm usually sympathetic to a [strangler vine pattern](http://www.martinfowler.com/bliki/StranglerApplication.html) gradual-adoption approach.  But here, I'm tempted to say that our frame is too new to need to be wrapping old growth in new growth vines so gingerly -- our frame *is* the new growth vines, and we should consider boldly adopting Material Design, biting the bullet and making the change now before we accumulate technical debt and existing usages sufficient to make this change prohibitively daunting.

smargovsky:
To help me ( and maybe the conversation ), could those for this change give specific examples where Material Design is better than ( Angular UI ) Bootstrap?  What would we gain / what's the potential?

I understand the ideas behind material design make it more encompassing than Bootstrap, and thus more strict on specific implementations of components / interactions / etc.  However, IMHO the actual framework ( http://www.getmdl.io/ and https://material.angularjs.org/latest/#/ ) doesn't strike me as better than Bootstrap, just different.  Thoughts?

apetro:
The difference I'm seeing is the aspiration to being an all-encompassing system for the user interface, rather than a bucket of UI components with the exercise left to the reader to figure out how to choose among them and how they relate.   I grant that that's a cartoonish simplification -- that Bootstrap is more than a bucket-of-loose-parts, and current Material Design resources are less than they aspire to be.  Nonetheless.  The all-encompassing system aspiration, if that were to pan out there'd be some upside to going along for the ride.

vertein:
Yeah!  I guess I don't really understand the visual differences to the user that well...  Some examples would be great of the differences or new features.  I'll admit that I'm not the biggest Material Design fan (and I have an Android), but I know that most people seem to like it.  Maybe there are benefits that are more than just visual to the user?

timlevett:
Checkout the [demo](https://material.angularjs.org/latest/#/demo/) area for examples of different directives. I agree with @apetro about the diff between bootstrap and material. However, I would also like to note that this doesn't mean we have to switch to be all material. We could just use some of there components, like [forms](https://material.angularjs.org/latest/#/demo/material.components.input) for example, or [sidenav](https://material.angularjs.org/latest/#/demo/material.components.sidenav).

This also opens doors for frame apps to have more options.

jhanstra:
Thoughts for when we discuss this in person:

Personally, I have to agree with @smargovsky and @vertein. I'm not sold on Material Design, or at least not this implementation of the idea of material design. I think the fact that it is so encompassing leaves little room for customization, and would lead to a very Google-y/Android-y look and feel. I think it's important for our MyUW brand not to be identifiable this way. Right now, you can't immediately tell that our site uses Bootstrap, and it retains a certain unique branding. If we used Material throughout, I think we would lose that branding and be bound by Material's rigid structure. The animations especially would be hard to override, and can be a bit over the top IMHO.

I also think to bite the bullet and implement this instead of Bootstrap would be a HUGE undertaking - not like a 20 or 40 point story but more of a 200 - 400 point story. And without a clear benefit and clear consensus on whether it's a good decision, that's pretty scary. If Material had come out a year ago, I would be much more welcoming to the idea of using it from the beginning, but now that our Beta is so stable and has generally gotten good reviews from users, I'm hesitant to completely revamp the UI with a new system. I'm still in favor of using certain components as we need them, though, and playing with it until we know whether or not we prefer it to Bootstrap/Angular Bootstrap.

timlevett:
Good thoughts @jhanstra . I am in no way suggesting we cut out bootstrap or what we have done. More just add this in. As you can see, this doesn't break anything due to class name difference, and the way we are defining our CSS. Our CSS trumps material. 

WRT customization, do we need to customize past coloring? They do have theme support, which is pretty easy to customize it a bit. A lot of the things we are doing are very similar to what they are doing. Circle with icon in middle that when you click does something, for one example. Instead of writing a custom solution we can use an out of the box solution.

We defiantly don't want to rewrite everything, that was not my intention of this, or my suggestion.

timlevett:
http://goo.gl/KT3SrA

nblair:
As much as I like Material, and I like the controls presented, I think it's probably too soon to introduce the style. Let Beta close, stick with Angular/Bootstrap styles and components for some time, then start to phase Material in larger swaths as a new "beta". Maybe liken it to Gmail and Inbox currently; with My UW on Bootstrap as Gmail, then a new frame built on Material as Inbox.

timlevett:
Good idea @nblair . We should at least wait until this is 1.x if we are going to utilize it. I'm going to close this PR. Great discussion everyone.  Thanks for participating.

PR #275, opened 2015-07-31
MUMUP-1999 : Remove beta flag on mobile toggle

timlevett:
So we can have widget be default, and for america.


vertein:
'murica :+1:

timlevett:
Party like its 1999

PR #277, opened 2015-08-03
EC: Cleanup some json for mock and beyond

timlevett:
+ Add expanded/compact to web.xml (404's are bad, mkay?)
+ Add mock data for earning statement widget
+ Change layoutDoc to use samples where available (courses and earning statement)
+ Fix json for new stuff (leading 0 be bad)
+ Fix unit tests


vertein:
:+1:

PR #278, opened 2015-08-03
MUMUP-1953   

jiayinjx:
Update a label to match updated branding of the CareerLocker external service.

### DIFF
--- a/angularjs-portal-frame/src/main/webapp/portal/main/partials/footer.html
+++ b/angularjs-portal-frame/src/main/webapp/portal/main/partials/footer.html
@@ -2,7 +2,7 @@
 	<div class="portal-power">
 		<div id="portalPageFooterLinks">
 			<div>
-				<span>© 2014, Board of Regents of the <a href="http://www.wisconsin.edu/" target="_blank">University of Wisconsin System</a></span>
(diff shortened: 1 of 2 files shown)


jhanstra:
:+1:

timlevett:
Please note that this doesn't satisfy the issue, just the mock data. You will have to change uPortal entity files in order to do this issue.

jiayinjx:
Not quite understand, could you explain a littler bit more?

in my test, it shows the label:CareerLocker instead of WISCareers, is that
what supposed to be?

timlevett:
The file you are modifying is just for the mock data, not for the real data in production. This data really comes from entity files that are managed in SVN on site, which you probably don't have access to. That is okay, but you should also update https://github.com/UW-Madison-DoIT/uPortal/blob/uw-master/uportal-war/src/main/data/uw_quickstart_entities/portlet-definition/wiscareers.portlet-definition.xml as part of this.

jiayinjx:
MUMUP-1971 manually update

timlevett:
Please create one PR per pull request, MUMUP-1971 and MUMUP-1953 are not related.

jiayinjx:
I was push then separately, don't know why they combined together.

PR #279, opened 2015-08-03
Update welcome image to reflect changes

jhanstra:
image


vertein:
Thoughts on changing the 'add to your homepage favorites' wording?  Maybe just 'add to home' or 'add to home page' to keep the top bubble consistent with the left bubble.

timlevett:
Agreed with the removal of favs

apetro:
Nitpick : "Switch between simple icons or informative widgets" 1) End with a period to be consistent with the other callouts.  2) I think "and" rather than "or" would sound more natural.

apetro:
I agree with @vertein and @timlevett that striking "favorites" from the text is desirable.

apetro:
Nuanced nitpick : The left hand bubble could be more about "Browse" and less about "find", that is, more like "Check out all the content in MyUW."  Whereas the tip bubble is already about "find".

jhanstra:
Good catches, will change

timlevett:
:cool: :+1:

vertein:
1 spot it's 'home page' and another it's 'homepage'

timlevett:
:+1: nice work @jhanstra . Merge?

PR #285, opened 2015-08-04
MUMUP-2017 cleanup some http.get calls that might not be configured in frame

timlevett:
+ Removes `ng-init` for welcome controller
+ Moves session service to create the promise later
+ Moves notification and group promise creations to later


apetro:
:+1:

PR #290, opened 2015-08-06
Accessibility - Header[MyUW logo]

jiayinjx:
issue description: pronounce weriod "myiuo" before.


vertein:
:+1:

timlevett:
I think this should be a constant that we could overwrite, just like `NAMES.title`

jiayinjx:
the reason I add the aria-label is that for screen reader, it reads out
weird like "my uiowa";
Keir told me he preferred the label be "my U W Home" instead of just "my U
W"
or do prefer aria-label="my U W"?

PR #293, opened 2015-08-07
MUMUP-2029 : Add in styling specificially for home

timlevett:
+ Added sidebar to our overlay project
+ Added `my-app.less` with just a simple class. We can refactor uw-frame and this later


jhanstra:
:+1: this is exciting

PR #295, opened 2015-08-07
improve the accessibility for widget view

jiayinjx:
could recognize all containers title on widget page now.
also skip the icon for not repeat read the {{::portlet.title}} again.


timlevett:
:+1:

PR #296, opened 2015-08-10
Accessibility improvement - search & browse MyUW page

jiayinjx:
Portlets' title able to get by TAB
the user will know which portlet it is before they "launch" it
the description of each app could able to read out.
"launch" described as "go to this app"


timlevett:
Please add the Jira key to the title

timlevett:
:+1:

PR #297, opened 2015-08-10
Accessibility improvement - apps/details/  page

jiayinjx:
Portlets' title able to get by TAB
the user will know which portlet it is before they "launch" it
make all of those four columns able to be read out in vertical order ( title -> content under the title)

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
@@ -47,9 +47,8 @@
 			      <span class="fa fa-plus"></span> Add to home
 			    </button>
 
-		    <sub class="sr-only" id="goTo-portlet.title-{{::portlet.fname}}">go to this app</sub>
             <!-- Launch button with access -->
-		    <a aria-labelledby="goTo-portlet.title-{{::portlet.fname}}" class="btn btn-success" href="{{::portlet.maxUrl}}" target="{{::portlet.target}}" ng-if="portlet.canAdd"><span class="fa fa-arrow-circle-o-right"></span> Launch</a>
+		    <a aria-label="launch {{::portlet.title}}" class="btn btn-success" href="{{::portlet.maxUrl}}" target="{{::portlet.target}}" ng-if="portlet.canAdd"><span class="fa fa-arrow-circle-o-right"></span> Launch</a>
 
 		    <!-- Launch button without access -->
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -47,8 +47,9 @@
 			      <span class="fa fa-plus"></span> Add to home
 			    </button>
 
-		    <!-- Launch button with access -->
-		    <a class="btn btn-success" href="{{::portlet.maxUrl}}" target="{{::portlet.target}}" ng-if="portlet.canAdd"><span class="fa fa-arrow-circle-o-right"></span> Launch</a>
+		    <sub class="sr-only" id="goTo-portlet.title-{{::portlet.fname}}">go to this app</sub>
```
apetro: How about `go to {{::portlet.title}}` rather than `go to this app`?  Seems like a worrying use of the term "app". :smile:

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -57,24 +58,24 @@
 	</div>
 	<div class="row no-margin"  ng-if='!loading && !error'>
 		<div class="col-xs-12 col-sm-4 col-lg-3 desc-item">
-			<h3 class="center">Description</h3>
-			<p class="app-description">{{::portlet.description}}</p>
+			<h3 tabindex="0" aria-label="Description" class="center">Description</h3>
```
apetro: This is fine if it's necessary, but it makes me worried about how poorly I understand accessibility, since i would have hoped the `Description` in the content of the `<h3` element would have been good enough to get a screen reader to understand that "Description" is the level-three heading here?

jiayinjx: That was what I thought before I tested it.
all the content could not be read out in the description. However,
sometimes, different screen reader have different ability. Like for JAWS,
we could use up and down arrow to let it read the content, but as a people
who could not know where is the header or content (which level it is), how
could he know where is the content. I am confusing about it, too.

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
jiayinjx: sounds great. Will do it.

timlevett:
Please add the Jira key to the title

timlevett:
:+1:

PR #299, opened 2015-08-11
MUMUP-2032 : add google events for categories

timlevett:
Note that the initial landing on the page won't trigger an event


vertein:
:+1: Tracking time.

jhanstra:
:+1:

PR #302, opened 2015-08-13
MUMMNG-1784 : Add in notification for user activation required

timlevett:
+ Move the MyUW specific notifications to angularjs-portal
+ Added in Activate Services notification 

Group came from the group the portlet is in. Text came from what the portlet currently says.


vertein:
:+1:

apetro:
:+1:

PR #309, opened 2015-08-24
Solve the accessibility effect for CSS 'display' rules

jiayinjx:
"display:none / visibility:hidden / overflow:hidden " have the accessibility effect that content is ignored by screen readers

### DIFF
--- a/angularjs-portal-home/src/main/webapp/css/home.less
+++ b/angularjs-portal-home/src/main/webapp/css/home.less
@@ -226,10 +226,15 @@ div.table-cell {
 .list-content .widget-remove,
 .widget-frame .widget-info,
 .widget-frame .widget-remove {
-  display:none;
+  display:inline;
+  position:absolute;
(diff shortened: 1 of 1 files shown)


timlevett:
@jhanstra - please check this out. Wouldn't this make it always show? even on non-hover state?

jhanstra:
Yeah, that would make it always show. I don't think we want that. Take a look at this article, @jiayinjx: http://alistapart.com/article/now-you-see-me. There are some suggestions for what to do for elements that should be available to screen readers. The last option in the table is probably the best option for this situation - `position: absolute; left:-999em`.

jiayinjx:
This is the same article I read. I tried to use position: absolute;
left:-999em., but after I switch "display : none" to it, the
'widget-remove' icon will disappear when hover.

timlevett:
:+1:

jhanstra:
:+1:

PR #313, opened 2015-08-27
MUMUP-2081 : create mock data for kv

timlevett:
+ Cleaned up some old test runners for codenvy
+ Created mock data for new key/value store


timlevett:
replaced with https://github.com/UW-Madison-DoIT/angularjs-portal/pull/314

PR #315, opened 2015-08-28
Moving storage kv to uw-frame as that is where notifications is

timlevett:


vertein:
get it out of here
:+1:

PR #316, opened 2015-09-01
MUMMNG-1524: Adds metric units to weather widget

Daha5:
Added a dropdown button that, when clicked, gives the user the choice for weather in imperial units or metric units. 

The code I added is in the weather widget controller (between lines 46-106 of controllers.js). The functions on lines 77 and 90 are called when the user clicks an item on the dropdown menu, and that does the calculations from Fahrenheit to Celsius, and from Celsius to Fahrenheit.

Temperature in degrees Fahrenheit
weather widget

Dropdown button is clicked.
weather widget2

Temperature is now in degrees Celsius
weather widget3

The reason the images are not showing is because I was running this locally on the angular-mock portal, and I think that's also why the temperature is getting cut off at the bottom.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
@@ -2,194 +2,189 @@
 
 define(['angular'], function(angular){
 
-	var app = angular.module('my-app.layout.widget.controllers', []);
+  var app = angular.module('my-app.layout.widget.controllers', []);
 
-	app.controller('OptionLinkController', ['$scope', 'layoutService', function($scope, layoutService){
+  app.controller('OptionLinkController', ['$scope', 'layoutService', function($scope, layoutService){
 
-		var configInit = function(){
-			if(!$scope.config) {
-				//setting up defaults since config doesn't exist
-				$scope.config = {
-						singleElement : false,
-						arrayName : 'array',
-						value : 'value',
-						display : 'display'
-				};
-			}
-		};
+    var configInit = function(){
+      if(!$scope.config) {
+        //setting up defaults since config doesn't exist
+        $scope.config = {
+            singleElement : false,
+            arrayName : 'array',
+            value : 'value',
+            display : 'display'
+        };
+      }
+    };
 
-		var populateWidgetContent = function() {
-			if($scope.portlet.widgetURL && $scope.portlet.widgetType) {
-				//fetch portlet widget json
-				$scope.portlet.widgetData = [];
+    var populateWidgetContent = function() {
+      if($scope.portlet.widgetURL && $scope.portlet.widgetType) {
+        //fetch portlet widget json
+        $scope.portlet.widgetData = [];
 
-				layoutService.getWidgetJson($scope.portlet).then(function(data) {
-					if(data) {
-						console.log(data);
-						//set init
-						if($scope.config.singleElement) {
-							//set the default selected url
-							$scope.portlet.selectedUrl = $scope.portlet.widgetData[$scope.config.value];
-						} else if($scope.portlet.widgetData[$scope.config.arrayName] && $scope.portlet.widgetData[$scope.config.arrayName].length > 0) {
-							$scope.portlet.selectedUrl = $scope.portlet.widgetData[$scope.config.arrayName][0][$scope.config.value];
-						}
-					} else {
-						console.warn("Got nothing back from widget fetch for " + $scope.portlet.fname);
-					}
-				});
-			}
-		};
-		configInit();
-		populateWidgetContent();
-	}]);
+        layoutService.getWidgetJson($scope.portlet).then(function(data) {
+          if(data) {
+            console.log(data);
+            //set init
+            if($scope.config.singleElement) {
+              //set the default selected url
+              $scope.portlet.selectedUrl = $scope.portlet.widgetData[$scope.config.value];
+            } else if($scope.portlet.widgetData[$scope.config.arrayName] && $scope.portlet.widgetData[$scope.config.arrayName].length > 0) {
+              $scope.portlet.selectedUrl = $scope.portlet.widgetData[$scope.config.arrayName][0][$scope.config.value];
+            }
+          } else {
+            console.warn("Got nothing back from widget fetch for " + $scope.portlet.fname);
+          }
+        });
+      }
+    };
+    configInit();
+    populateWidgetContent();
+  }]);
 
-	app.controller('WeatherController', ['$scope', 'layoutService', function($scope, layoutService){
-		$scope.weatherData = [];
-		$scope.loading = false;
-		$scope.showMetric = false;
-		$scope.tempInC1 = 0;
-		$scope.tempInC2 = 0;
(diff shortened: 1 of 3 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
```diff
@@ -1,3 +1,14 @@
+
+    <div class="weather-dropdown" dropdown is-open="status.isopen">
+      <button id="single-button" type="button" class="btn btn-primary rounded" dropdown-toggle ng-disabled="disabled">
+       <i class="fa fa-wrench "></i>
+      </button>
+      <ul class="dropdown-menu" role="menu" aria-labelledby="single-button">
+        <li role="menuitem"><a href="" ng-click="showMetric=false; changeToF()">Imperial</a></li>
```
apetro: Looks like ["Imperial" may not be the best way to refer to this](https://en.wikipedia.org/wiki/United_States_customary_units).  Consider instead "US customary" or maybe even just get right to the point and call these "Fahrenheit" and "Celsius", even if they imply different presentation of things other than temperature too.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -2,163 +2,194 @@
...
+						}
+					} else {
+						console.warn("Got nothing back from widget fetch for " + $scope.portlet.fname);
+					}
+				});
+			}
+		};
+		configInit();
+		populateWidgetContent();
+	}]);
```
vertein: It would look like your text editor (probably eclipse?) is inserting tabs for spaces.  Not a big deal, but when you get a chance, go ahead and change that to insert spaces for tabs.  If you want, I'll show you what I'm talking about and where to do it.

apetro:
Should this be using the within-weather-portlet user preference about units?

within-portlet units selector

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -2,163 +2,194 @@
...
+					$scope.weatherData[i].currentWeather.temperature = ($scope.weatherData[i].currentWeather.temperature)*(9/5) +32;
+					
+					for (var j=0; j <$scope.weatherData[i].forecast.length; j++){
+						$scope.weatherData[i].forecast[j].highTemperature = ($scope.weatherData[i].forecast[j].highTemperature)*(9/5) + 32;
+						$scope.weatherData[i].forecast[j].lowTemperature = ($scope.weatherData[i].forecast[j].lowTemperature)*(9/5) + 32;
+					}
+				}
+				$scope.currentlyImperial = true;
+			}
+		}
```
timlevett: Curious if there is a way we just have a "toggle" function instead of having 2 functions that do pretty much the same thing.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -2,163 +2,194 @@
...
-        };
+	app.controller('WeatherController', ['$scope', 'layoutService', function($scope, layoutService){
+		$scope.weatherData = [];
+		$scope.loading = false;
+		$scope.showMetric = false;
+		$scope.tempInC1 = 0;
+		$scope.tempInC2 = 0;
+		$scope.celsiusForecasts = [];
+		$scope.dailyHigh = 0;
+		$scope.dailyLow = 0;
```
timlevett: What are all these used for?

timlevett: Looks like tempInC1, tempInC2, celsiusForecasts and dailyhigh/low are not used

timlevett:
@apetro - funny enough, regardless what you put there, it comes back with F

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
jhanstra: Also in favor of 'Fahrenheit' and 'Celsius'

timlevett: Third-ed.

jhanstra:
UI-wise, the red button is pretty distracting and asymmetrical. Thoughts on making it like the info button icon and remove icon so it's a little more subtle and tucked away?

jhanstra:
Ahh, didn't mean to close it, dangit Github

apetro:
Is this the moment to invent an end-user-facing `/settings` page, now that we've got a setting?

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
```diff
@@ -26,13 +37,16 @@
                         <div class="col-xs-3 day">
                           <p>Now</p>
                           <img ng-src="{{config.iconPrefix}}{{weather.currentWeather.imgName}}{{config.iconPostfix}}" title="{{weather.currentWeather.condition}}" alt="{{weather.currentWeather.condition}}">
-                          <p>{{weather.currentWeather.temperature}}&deg; {{weather.temperatureUnit}}</p>
+                          <p ng-show="!showMetric" >{{weather.currentWeather.temperature | number:0}}&deg; {{weather.temperatureUnit}}</p>
```
apetro: Idea: what if the temperature unit ( "C" or "F") *were* the toggle to switch between units?  Don't like degrees F?  Click the F and switch it to C.

Would be way, way more subtle.  Maybe too subtle.  Some users wouldn't notice it.  But then most users accept the default units, and anyone who cares enough to raise it to the helpdesk could be walked through clicking the unit to change it.

timlevett:
@apetro - interesting proposal about the global settings page. Brained on it a bit, came up with this: https://gist.github.com/timlevett/86cbde0dd60a327964cf

timlevett:
Another thought is having that settings mode of each card. So you have a little gear next to the info, when you click the card flips to config mode where you can configure that widget. Might be better.

apetro:
@timlevett as I commented there, that generated-from-widget-metadata settings page feels *brilliant*.  Tremendous potential.  I really, really like it.

I'm just a tiny bit concerned that it's too much too soon, that we should circle back to refactor to that when we have three or four settings.  Today's problem is just one setting.  Maybe we don't quite yet need a framework for settings, as awesome as such a framework will be.  Today we just need a setting?  Maybe that's a reason to do the just-make-the-unit-letter-clickable solution, not because that solution is awesome, but because that solution defers inventing a bigger solution?

timlevett:
Agreed @apetro, just future thought really. No need to make a global settings feature for one little toggle :)

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
timlevett: The call to action might be weird. What about just displaying both at the same time?

Daha5:
See https://github.com/UW-Madison-DoIT/angularjs-portal/pull/319

PR #317, opened 2015-09-02
Updates keyValueUrl to one that exists

vertein:


angularjs-portal-home/src/main/webapp/js/app-config.js
```diff
@@ -25,7 +25,7 @@ define(['angular'], function(angular) {
             },
             'notificationsURL' : '/web/samples/notifications.json',
             'groupURL' : '/portal/api/groups',
-            'kvURL' : '/portal/api/keyvalue/'
+            'kvURL' : '/storage/'
```
timlevett: This was more for mock data than anything else, hmmm, how to handle this... We do overlay it in environments.

vertein: could leave it null and overlay it?

timlevett: We could, or just leave it as `/portal/api/keyvalue` for mock dataness

vertein: Ahhh, I just looked.  I didn't realize there was actual mock data there.  I'll close this and we can overlay.

timlevett: :+1: for action taken

PR #318, opened 2015-09-03
Misc Cleanup

timlevett:
+ MUMMNG-1812 : Cleanup controller filter for widgets. There was a weird state of the filter (going to be used for time & absence portlet) where the service hasn't returned yet but we were trying to run the filter on an undefined object. It would vomit errors in the log. This checks for null first now. :)
+ MUMUP-2090 : Add generic error page in mock data with a redirect link to `/web`
+ EC : Add directory widget to `layoutDoc`


vertein:
:+1: "real portal"

PR #319, opened 2015-09-03
MUMMNG-1524: Adds metric units to weather widget - UPDATED

Daha5:
I think I broke the old pull request for this, so I created another one.

Continuation from https://github.com/UW-Madison-DoIT/angularjs-portal/pull/316

Added a dropdown button that, when clicked, gives the user the choice for weather in imperial units or metric units. 

The code I added is in the weather widget controller (between lines 46-106 of controllers.js). The functions on lines 77 and 90 are called when the user clicks an item on the dropdown menu, and that does the calculations from Fahrenheit to Celsius, and from Celsius to Fahrenheit.

Temperature in degrees Fahrenheit
weather widget

Dropdown button is clicked.
weather widget2

Temperature is now in degrees Celsius
weather widget3

The reason the images are not showing is because I was running this locally on the angular-mock portal, and I think that's also why the temperature is getting cut off at the bottom.

**UPDATED SCREENSHOTS:**

User now has the ability to toggle between Celsius and Fahrenheit
updated weather widget

Button fades out if the user tries to click Celsius when they're already viewing the temperature in Celsius.
updated weather widget 2

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
@@ -2,194 +2,189 @@
 
 define(['angular'], function(angular){
 
-	var app = angular.module('my-app.layout.widget.controllers', []);
+  var app = angular.module('my-app.layout.widget.controllers', []);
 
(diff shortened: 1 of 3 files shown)


timlevett:
:+1:

PR #321, opened 2015-09-03
MUMMNG-1857 : Adding more features to rss feed widget

timlevett:
Added the following configuration options for `<rss>`
+ `dateFormat` : The date format, see https://docs.angularjs.org/api/ng/filter/date
+ `showShowing` : show the Showing x of y (default `false`)

Here is an example with the `showShowing` off and a date with format `M/d`
image


Daha5:
Looks good :+1:

vertein:
:+1:

PR #323, opened 2015-09-04
Fix final bugs for the weather widget units change

Daha5:
Fahrenheit (F) button no longer stays gray when it's not clicked, and the buttons no longer overlap or get buggy when the widget width changes.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/css/widget.less
+++ b/angularjs-portal-home/src/main/webapp/css/widget.less
@@ -47,14 +47,27 @@ weather {
     }
   }
   .weather-dropdown{
-    position: absolute;margin-left:200px; margin-top: -42px;
+    position: absolute;
+    margin-top: -40px;
+    right:30px;
     }
   .weather-dropdown a{
-    margin: 0px 5px; color: #b70101;
+    margin: 0px 5px;
+     color: #b70101;
     }
   .weather-clicked{
     color:#b70101;
     }
+  @media (max-width: 1246px){
+    .weather-dropdown{
+      right:20px;
+    }
+  }
+  @media (max-width: 1200px){
+    .weather-dropdown{
+      right:30px;
+    }
+  }
 }
 .widget-content option-link {
   .widget-icon-container i {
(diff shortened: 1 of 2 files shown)


timlevett:
does this also fix the buttons overlapping the (x) button?

keirserrie:
Ahad--sorry to weigh in so late, but I'd love to have the interface be more subtle on this, like:

https://www.dropbox.com/s/jrsdydpcp261tzs/Screenshot%202015-09-08%2011.58.00.png?dl=0

Otherwise we could have more requests for buttons in the widget title (yikes)

timlevett:
@keirserrie - I like that thought, but what of the fat thumb'd peoples... #touchsurfaces

keirserrie:
@timlevett - good point, not ideal...already an issue with world weather online link, fat thumbs will press launch app. Could widen target by having one link:

Change to C°
or
Change to F° 

Could also just bury in the app interface and drop from the widget. (Sorry Ahad...)

keirserrie:
https://www.dropbox.com/s/6zxzdmw36miatc6/Screenshot%202015-09-08%2013.44.59.png?dl=0

timlevett:
nice @keirserrie

apetro:
From a product management perspective, I like a more subtle solution here, in that it sets fewer expectations and less precedent, leaves more freedom to continue to evolve how widget settings works in the future.

Daha5:
When the widget is at it smallest width, there wouldn't be enough room to display the entire "Change to C" (or "Switch to C") if it's on the same line as the world weather online link

weather widget no room

keirserrie:
API terms allow "Powered by World Weather Online," so if we do that and make "Change to C" same text size, should fit.

screenshot 2015-09-09 14 42 19

Daha5:
@keirserrie How about your original proposal?
weather widget more room

Daha5:
Final result looks like this: 
final result

timlevett:
:+1:

keirserrie:
:+1: awesome ahad

PR #325, opened 2015-09-13
MUMUP-1862 Marketplace categories bookmarkable

jiayinjx:


timlevett:
:+1: for code (tested locally). Does this make categories bookmarkable tho? I do agree that it makes you able to link back/forth through details and full marketplace list, but not sure if you would be able to place a bookmark on a category.

PR #327, opened 2015-09-15
MUMUP-1928 Grade&grading policy link

jiayinjx:
update the Grades and Grading policy link, it returned a 404 error (http://registrar.wisc.edu/error404.htm) before.


timlevett:
Note that this is an entity file change. Its nice that we are updating the mock data to reflect the change as well, but just know the story is not complete when doing these changes. Might be better to skip these issues until you have access to uPortal and entities.

timlevett:
:+1:

PR #328, opened 2015-09-17
MUMUP-2071 Hide "add more to home" when read-only(guestMode)

jiayinjx:
when the username is "Guest", Both the "add more to home" button on the home_header and container in icon/widget View will be hidden

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
@@ -221,12 +221,12 @@ define(['angular', 'jquery'], function(angular, $) {
             //Check guestMode
             var that = this;
             that.user = [];
-            $scope.notGuestMode = true;
+            $scope.GuestMode = false;
 
             mainService.getUser().then(function(result){
               that.user = result;
               if (that.user.displayName === "Guest")
-                  $scope.notGuestMode = false;
+                  $scope.GuestMode = true;
             });
 
         }]);
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
@@ -27,6 +27,7 @@ define(['angular', 'jquery'], function(angular, $) {
         '$rootScope',
         'layoutService',
         'miscService',
+        'mainService',
         'sharedPortletService',
         function($location,
                  $localStorage,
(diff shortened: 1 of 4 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -105,6 +107,17 @@ define(['angular', 'jquery'], function(angular, $) {
                 }
             };
 
+            //Check guestMode
+            var that = this;
+            that.user = [];
+            $scope.notGuestMode = true;
```
timlevett: I would prefer to have `$scope.guestMode` to remove the possibility of a double negative of not not being a guest ;)

angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -203,6 +218,17 @@ define(['angular', 'jquery'], function(angular, $) {
...
+
+            mainService.getUser().then(function(result){
+              that.user = result;
+              if (that.user.displayName === "Guest")
+                  $scope.notGuestMode = false;
+            });
+
```
timlevett: weird this have to be in 2 places. Suggestion of `$rootScope`?

angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-header.html
```diff
@@ -13,16 +13,22 @@
       </div>
     </div>
 </div>
-<div class="inner-nav-container">
+<div class="inner-nav-container" ng-controller="LayoutController as layoutCtrl">
   <ul class="inner-nav">
     <li ng-controller='GoToAppsController as GoToAppsCtrl'>
       <button
+       ng-show='!GuestMode'
```
timlevett: `ng-hide='GuestMode'`

angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-header.html
```diff
@@ -13,16 +13,22 @@
       </div>
     </div>
 </div>
-<div class="inner-nav-container">
+<div class="inner-nav-container" ng-controller="LayoutController as layoutCtrl">
```
timlevett: We don't want to load this controller twice if we don't have to. Is there a way we could avoid this?

timlevett:
I believe this is going to be reworked now that guest mode exists in frame.

PR #330, opened 2015-09-28
MUMMNG-1973 MUMMNG-1520 : tweak ABE notification text

apetro:
Updates text of Annual Benefits Enrollment notification to use "Annual Benefits Enrollment" vs "open enrollment" and not use "th" on dates, per Service Center style preferences.


timlevett:
:+1:

PR #331, opened 2015-09-29
MUMUP-2057 : Static controller now stops loading

timlevett:
Fixed loading gif staying if you click on the link to the app from home screen


vertein:
:+1:

PR #333, opened 2015-10-08
MUMUP-2102 Loading state flag feature for generic widgets

jiayinjx:


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -29,25 +29,28 @@ <h4 id="appTitle_portlet.title-{{::portlet.nodeId}}" aria-labelledby="appTitle_p
...
   <div ng-if="'RSS' === widgetCtrl.portletType(portlet)">
     <rss app="portlet" config="portlet.widgetConfig"></rss>
   </div>
-  
+
   <div ng-if="'GENERIC' === widgetCtrl.portletType(portlet)">
     <div ng-controller="GenericWidgetController as genericWidgetCtrl">
+        <div ng-if="loading" id="loading">
+            <loading-gif data-object='content' data-empty='isEmpty'></loading-gif>
+        </div>
```
timlevett: The original idea of the "loading state" was that the widget would have the option to have a loading state. I do like this uniformed approach tho. Makes way more sense.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -29,25 +29,28 @@ <h4 id="appTitle_portlet.title-{{::portlet.nodeId}}" aria-labelledby="appTitle_p
...
         <content-item></content-item>
```
timlevett: Should we add in a `ng-if='!loading'` here?

jiayinjx: I think we need it, when I try to remove it and run, the loading-gif will
be there forever.

timlevett: I was commenting saying we should add the !loading to content-item. So during loading it doesn't show the content-item content.

jiayinjx: got it, update just now.

timlevett:
:+1:

PR #334, opened 2015-10-09
MUMUP-2155 : Friendly zero search results

apetro:
When the user searches and finds zero search results, 

 * Clearly state that no results matched
 * Suggest ways to recover

Suggestions implemented in this changeset:

 * dropping the category filter, iff matching results are hidden by the category filter, 
 * showing non-launchable matches,  iff matching results are hidden by the "Show all" filter
 * searching in the directory, with a link that carries along the active query string
 * searching in UW Madison web search, with a link that carries along the active query string
 * searching in the KB, with a link that carries along the query string
 * asking the Helpdesk for help
 * submitting feedback about MyUW search

## Configurable and degrades gracefully

The external search URLs are configured in `app-config.js`, in that

 * Directory search URL is configurable; if not set this option is not shown.
 * Web search URL is configurable; if not set this option is not shown.  The UI label for the domain over which this searches is also configurable; if not set it just says Web.
 * KB search URL is configurable; if not set this option is not shown.

Honors pre-existing feedback URL configuration; if not set that option is not shown.  Adds new helpdesk URL configuration; if not set that option is not shown.

# Screenshots

## Basic suggestions

search for vertein no resutls

## When Show all would help, suggest it

google analy

## When All Categories would help, suggest it

suggest looking beyond the selected category

(this screenshot is messed up somehow, but this feature does work.)

## When no domain is configured for web search

no search domain- no worries

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
@@ -37,6 +37,23 @@
     ng-show="(portlets | filter:searchTermFilter | showApplicable:showAll | showCategory:categoryToShow).length == 0" 
     class="no-results">
     <p>No matching MyUW content.</p>
+    <p>Maybe try:</p>
+    <ul>
+      <li>
+        <a href="http://www.wisc.edu/directories/"
+           target="_blank">Search the directory</a> (of people and offices) instead.</li>
+      <li>
+        <a href="http://www.wisc.edu/search/"
+           target="_blank">Search the wisc.edu domain on the Web</a> instead.</li>
+      <li>
+        <a href="https://kb.wisc.edu/"
+           target="_blank">Search the KnowledgeBase</a> instead.</li>
+      <li>
+        <a href="https://kb.wisc.edu/helpdesk/"
+           target="_blank">Ask the Helpdesk</a> for help.</li>
+      <li>
+        <a href="/portal/p/feedback">Give feedback</a> about search.</li>
+    </ul>
   </div>
   
   <rating-modal-template></rating-modal-template>
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
@@ -50,8 +50,8 @@
       <li ng-if="webSearchUrl">
         <a ng-href="{{webSearchUrl}}{{searchText}}"
            target="_blank">Search the wisc.edu domain on the Web</a> instead.</li>
-      <li>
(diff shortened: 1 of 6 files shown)


angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
```diff
@@ -32,5 +32,34 @@
...
+        <a ng-href="http://www.wisc.edu/search/?q={{searchText}}"
+           target="_blank">Search the wisc.edu domain on the Web</a> instead.</li>
+      <li>
+        <a ng-href="https://kb.wisc.edu/search.php?q={{searchText}}"
+           target="_blank">Search the KnowledgeBase</a> instead.</li>
+      <li>
+        <a href="https://kb.wisc.edu/helpdesk/"
+           target="_blank">Ask the Helpdesk</a> for help.</li>
+      <li>
+        <a href="/portal/p/feedback">Give feedback</a> about search.</li>
```
timlevett: There is a MISC_URL for this.

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
```diff
@@ -32,5 +32,34 @@
...
+        ng-show="(portlets | filter:searchTermFilter | showCategory:categoryToShow).length > 0">
+         <a href="#" ng-click="toggleShowAll()">Show matching MyUW content even though it is not launchable</a>.</li>
+      <li>
+        <a ng-href="http://www.wisc.edu/directories/?q={{searchText}}"
+           target="_blank">Search the directory</a> (of people and offices) instead.</li>
+      <li>
+        <a ng-href="http://www.wisc.edu/search/?q={{searchText}}"
+           target="_blank">Search the wisc.edu domain on the Web</a> instead.</li>
+      <li>
+        <a ng-href="https://kb.wisc.edu/search.php?q={{searchText}}"
```
timlevett: It would be good to have these URLs in MISC_URLs, or maybe a new contant group. Thinking of master implementer..

timlevett:
Overall thought is :+1: for improved 0 state experience. Good easy win. Could be better for other adopters. Noted some comments inline with suggestions about using `app-config.js`.

timlevett:
Before putting any more effort into this should we wait for the design sprint for search to be completed?

jhanstra:
:+1:

apetro:
I figure this makes the existing no-search-results experience better, more user recoverable, as a quick win, and then following up on the search design sprint, search can continue to get more better.

apetro:
Now configured via `MISC_URLS`, with graceful degradation. Adopters can support many or few no-search-results recovery options, and can even configure away the "wisc.edu" characterization of the Web search option.

keirserrie:
This is beautiful. Probably want to rejigger the wording a bit at some point (e.g., "domain").

PR #336, opened 2015-10-14
Only show no-results content after portlet registry loads.

apetro:
Only show the no-search-results recovery content after the array of portlets available for filtering is non-empty.

While loading:

now with clever loading goodness

After loading:

loaded

Tested locally using Chrome dev tools throttling, which is pretty cool.


apetro:
Let's hold off on merging this, I think I can make this better...

timlevett:
:+1:

PR #337, opened 2015-10-15
MUMUP-2163 Add events search as a recover option

apetro:
Adds "Search for events instead" as another no-search-results recovery option.

This tweak is intended to be responsive to a request around content in MyUW supporting the [AAU Survey Student Engagement Sessions](https://today.wisc.edu/events/search?utf8=%E2%9C%93&term=AAU+Survey+Student+Engagement+Session&commit=Search) without committing MyUW to adding a `portlet-definition`, widget, entry in the directory of apps, notification, etc., supporting these three point-in-time events.

The specific zero-search-results search that raised the concern was a search for "sexual assault" and concern that this yields no search results. The compromise initial response is "we're not able to immediately make that search term directly yield useful content, but if people were searching for this event, they'll see a search recovery option that looks like a next step for looking for an event."

search for events instead


timlevett:
:+1:

PR #339, opened 2015-10-22
MUMMNG-1983 Rename FacStaff group

jiayinjx:
Rename "FacStaff" which include users neither faculty nor staff, to "users-with-a-UDDS-number", reflect what is really is.


timlevett:
Note that this is not the complete solution. There are changes to be made in uPortal for this group.

timlevett:
Reviewing the issue, this would work, but this story was not groomed yet and didn't have a suggested path. The path you took here was a possible solution, but there is a simpler way. I'm going to close this PR. In the future, please work with the dev team on tasks that are already groomed and ready for development.

PR #345, opened 2015-11-03
[WIP] Use directive for home

timlevett:
+ Comment out current header
+ Use `portlet-header` directive, smaller size, with variable background, no toggle
http://goo.gl/qTJFvn


keirserrie:
:+1: Nicely done, I still need to figure out what goes in that header

timlevett:
@keirserrie thanks! I did it more as a conversation starter. What do we really want.  Also did this for fun: http://goo.gl/s4al4p

keirserrie:
Good to have options...like the idea of no header on home, but may need to rethink the top part. Good stuff

timlevett:
Chatted with Keir, lets merge this and see what happens.

apetro:
:+1: It's an improvement.  Merging this makes it better now, and it can be made more better later.

PR #347, opened 2015-11-06
use ng-switch for widget selection

timlevett:
Read about `ng-switch` and thought this would be a good case to use it. Did a little logging, for someone that has a single widget on there page, it reduces the amount of calls to `widgetCtrl.portletType(portlet)` from 188 to 42. Decent improvement.


vertein:
I do love me some ng-switch.  :+1:

PR #349, opened 2015-11-09
MUMMNG-2069 Add search keywords for Benefits Information

jiayinjx:


timlevett:
These changes are for entities changes. A good frame of reference, if you are changing anything in `angularjs-portal-mock-portal` then its something that needs to be changed in [uPortal](https://github.com/UW-Madison-DoIT/uPortal). mock data is only used by us developers and doesn't really matter if its out of date.

PR #350, opened 2015-11-10
Filter on home

timlevett:
AngularJS comes stock with really neat string filtering. 

As a "heavy" user of MyUW, I found if I have more than 10 widgets on my home screen, things are hard to find quickly w/o doing a CTRL+F and searching for the string. Maybe that is just the CS person in me? 

SO, I thought I would try adding in an alpha flag that lets me have my ctrl+f.

+ Add alpha settings page for /web (under `/web/settings/web`)
+ Adds flag for filter
+ Adds filter for widget and list view

#### New alpha settings page
http://goo.gl/3eZvnA

#### Filter Button before click

http://goo.gl/ox6uaA

#### Filter After Click of Filter

http://goo.gl/nPKLRo

#### Animation of a filter 

http://goo.gl/rrQ3Up


apetro:
:+1: .  If/when this merges let's tee up a Story to get some user testing time on this, so that we can decide to :shipit: for real, not-hidden-by-flag.  But I love the flag approach initially to allow user testing it for reals in prod without committing to the feature yet.

angularjs-portal-home/src/main/webapp/my-app/layout/list/partials/home-list-view.html
```diff
@@ -1,10 +1,13 @@
 <div class="row portlet-frame">
   <home-header></home-header>
   <div role="main" class="col-xs-12 no-padding portlet-body" ng-controller="LayoutController as layoutCtrl">
+    <div class='row col-xs-12' ng-show='showHomeFilter' style='margin: 5px; margin-left: 0;'>
+      <input ng-model="searchText" placeholder='Filter' class='form-control'>
```
apetro: Maybe call this something other than `search`. `filterText` maybe?  Search makes it too easy to confuse with the top :mag: search bar :confused: .

pom.xml
```diff
@@ -244,7 +244,7 @@
                 </goals>
               </execution>
             </executions>
-          </plugin>
+          </plugin>-->
```
apetro: Unrelated change?

timlevett: Heh, opps. Yep. That plugin doesn't work in codenvy, thought i reverted that before committing. I'll remove this change.

timlevett: This is why we have code reviews :)

angularjs-portal-home/src/main/webapp/my-app/layout/list/partials/home-list-view.html
timlevett: Its just the variable name?

timlevett: But sure, i'll change it to filterText

timlevett:
Chatted with @keirserrie . Going to commit as its hidden anyway so he can play.

keirserrie:
Here's an idea--start with buttons w/explicit text:

filter-01 002

Pressing filter button takes over the entire area:

filter-01 003

timlevett:
nice, we could even do a cute transition or something flashy.

PR #351, opened 2015-11-11
Make list-of-links widget type more responsive

jhanstra:
Three things were done:
1. Remove bootstrap padding on columns in list-of-links widgets
2. Define height of 95px
3. Truncate after 24 characters (code change in uw-frame)

Before:
image

After:
image

### DIFF
--- a/angularjs-portal-home/src/main/webapp/css/widget.less
+++ b/angularjs-portal-home/src/main/webapp/css/widget.less
@@ -93,4 +93,8 @@ weather {
     color:@color1;
     padding-right:10px;
   }
+  .col-xs-6 {
+    height:95px;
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/css/widget.less
```diff
@@ -93,4 +93,8 @@ weather {
     color:@color1;
     padding-right:10px;
   }
+  .col-xs-6 {
+    height:95px;
+    padding:0px !important;
+  }
```
timlevett: Dig this because its wrapped in `lol` tag

timlevett:
:+1:

PR #355, opened 2015-11-23
Cleanup settings page

timlevett:
+ Removes app specific beta settings
+ Adds in frame based beta settings config (see pull for uw-frame [here](https://github.com/UW-Madison-DoIT/uw-frame/pull/75))

image


vertein:
:+1: thanks frame.

PR #357, opened 2015-11-26
Brand Help Desk correctly (rather than Helpdesk).

apetro:
The no-search-results recovery link about asking for help didn't match the Help Desk's [branding as Help Desk](https://kb.wisc.edu/helpdesk/). This fixes that.


timlevett:
:+1:

PR #366, opened 2016-01-04
MUMUP-2153 : Render launched simple content under /web/static

timlevett:
+ Replaced mock data with up to date api
+ added URLs to `/web/static/fname` for `cms`


vertein:
:+1: looks good

PR #374, opened 2016-01-25
MUMUP-2199 : refactor search to be in search package

timlevett:
+ Moved search content to search package.
+ Added search service. 
+ Added Google Search function. 
+ Added Framing for search result controller


vertein:
I don't normally :+1: something that failed integration checks, but I'll make an exception.

vertein:
:+1: on condition that checks pass

PR #386, opened 2016-01-29
MUMUP-2357 : update to have background style, for the style

timlevett:
Need this first: https://github.com/UW-Madison-DoIT/uw-frame/pull/138 :white_check_mark: 

+ adds background to static
+ adds background to exclusive

before
http://goo.gl/vro8Ft

after
http://goo.gl/3NhBr4

Good catch @Daha5


vertein:
:+1:

jhanstra:
If we're going to have margins on the right and left to give the idea of a background, we might want to also have a margin on the top. It looks sort of odd to me as is. Thoughts @keirserrie?

PR #389, opened 2016-02-02
Filter search results

jhanstra:
Filters work on the search results page. I used jQuery to do this, because it seemed like the easiest option.

Room for improvement: 
- Loading state. Initial load takes 5-7 seconds to get results, but doesn't show the user that more results are coming. 
- We might want to show a 'No results' paragraph for each section, so that when you filter it to, say, MyUW results and there are none, it isn't blank.

We can try to get these improvements in before releasing tomorrow, but I think we should at least get this in.

screen recording 2016-02-02 at 02 44 pm

### DIFF
--- a/angularjs-portal-home/src/main/webapp/css/search-results.less
+++ b/angularjs-portal-home/src/main/webapp/css/search-results.less
@@ -8,6 +8,11 @@
   hr {
     margin: 0;
   }
+  .inner-nav-container .inner-nav li {
+    &:focus {
+      outline:0;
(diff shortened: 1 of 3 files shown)


timlevett:
:+1: Like the iterative approach.

apetro:
This doesn't feel Done without a bit more polish on handling loading-indication and empty states.

timlevett:
I have a quick fix for the loading states, I'll add a pr once this is merged

PR #390, opened 2016-02-02
MUMUP-tbd : Happy loading, Happy no results

timlevett:
Loading state is too fast, but all sections have a loading gif.

http://goo.gl/uhGdUM

PR #391, opened 2016-02-02
Tweaks on search results polish

apetro:
Grab bag of tweaks to the pending #390 . Let's take the ones that seem to help.

# Show loading indicator in place of results count when loading

It's kind of subtle, but here's the gif

loading indicator in search tab bar

# Suppress no-results search results sections on All tab

Show the search sections either when

1. They have results, or
2. They are the focused-upon search section.

That is, don't show them on the `all` tab when they have no results, since their presence in the search tabs is enough to let the user know what kinds of results had none.

suppress empty results on all tab

# Show the inline heading and `<hr>` only on the All tab

So, 

show heading on all tab

but

suppress heading

# Suggest browse as a way to recover from no-results MyUW search

suggest browsing to recover from no-results myuw search

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
@@ -48,6 +48,7 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       var initWiscEduSearch = function(){
         googleCustomSearchService.googleSearch($scope.searchTerm).then(
           function(results){
+            $scope.googleSearchLoading = false;
             if(results && results.responseData && results.responseData.results) {
               $scope.googleResults = results.responseData.results;
               if(results.responseData.cursor.estimatedResultCount){
@@ -61,6 +62,7 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       var initWiscDirectorySearch = function(){
         wiscDirectorySearchService.wiscDirectorySearch($scope.searchTerm).then(
           function(results){
+            $scope.directorySearchLoading = false;
             if(results){
               if(results.records && results.count) {
                 $scope.wiscDirectoryResults = results.records;
@@ -139,9 +141,11 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       };
       init();
       if(googleCustomSearchService.googleSearchEnabled()){
+        $scope.googleSearchLoading = true;
         initWiscEduSearch();
       }
       if(wiscDirectorySearchService.wiscDirectorySearchEnabled()){
+        $scope.directorySearchLoading = true;
         initWiscDirectorySearch();
       }
     }]);
@@ -149,4 +153,3 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
     return app;
 
 });
-
--- a/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
@@ -84,21 +84,25 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
           $('#myuw-results').show();
           $('#wisc-directory-results').show();
           $('#wisc-edu-results').show();
+          $scope.activeFilter = "all";
         } else if (filterName == 'myuw') {
           $('#myuw-selector').addClass('active');
           $('#myuw-results').show();
           $('#wisc-directory-results').hide();
           $('#wisc-edu-results').hide();
+          $scope.activeFilter = "myuw";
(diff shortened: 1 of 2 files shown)


timlevett:
I merged my pr (the one this is based off) into master (2 :+1: from the team). You can redirect this one toward master.

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
```diff
@@ -5,23 +5,28 @@
...
-    <hr>
+  <div id="myuw-results" class='search-results-container'
+    ng-show="myuwFilteredResults.length > 0 
+    || activeFilter == 'myuw' ">
+    <h4 class="header" ng-show="activeFilter == 'all'">MyUW</h4>
+    <hr ng-show="activeFilter == 'all'">
     <loading-gif data-object='myuwResults'></loading-gif>
     <div ng-show="myuwFilteredResults.length == 0" class='no-result'>
-      No MyUW results
+      No MyUW results. <a href="/web/apps">Try browsing instead?</a>
```
timlevett: This can be just apps (base = /web)

angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
```diff
@@ -134,9 +141,11 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       };
       init();
       if(googleCustomSearchService.googleSearchEnabled()){
+        $scope.googleSearchLoading = true;
```
timlevett: Move these loading flags into the init function right before the $http.get. Having it here takes it out of context and its confusing.

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
```diff
@@ -5,23 +5,28 @@
         <a>All ({{ totalCount }})</a>
       </li>
       <li ng-click="filterTo('myuw')" id="myuw-selector">
-        <a>MyUW ({{myuwFilteredResults.length}})</a>
+        <a>MyUW <span ng-hide="portletListLoading">({{myuwFilteredResults.length}})</span>
+        <i ng-show="portletListLoading" class="fa fa-spinner fa-pulse"></i></a>
```
timlevett: Interesting design decision. Do we think 2 loading gifs for the same section may be too much (one in the body, one in the tab)? @keirserrie ? I personally could go either way, maybe i'd have to see it in slow mo.

angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
```diff
@@ -84,21 +86,25 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
           $('#myuw-results').show();
           $('#wisc-directory-results').show();
           $('#wisc-edu-results').show();
+          $scope.activeFilter = "all";
```
timlevett: Doing this makes the `jquery` implementation redundant. I suggest choosing one and go with it.

angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
```diff
@@ -61,6 +62,7 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       var initWiscDirectorySearch = function(){
         wiscDirectorySearchService.wiscDirectorySearch($scope.searchTerm).then(
           function(results){
+            $scope.directorySearchLoading = false;
```
timlevett: Make sure to set this to false in the error block also. If it fails, then it would load forever (in its current state)

angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
```diff
@@ -48,6 +48,7 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
       var initWiscEduSearch = function(){
         googleCustomSearchService.googleSearch($scope.searchTerm).then(
           function(results){
+            $scope.googleSearchLoading = false;
```
timlevett: Set this in the error block of the promise as well

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
timlevett: Additionally we should probably be consistent with the flags we are using for loading state. Currently we are checking both the length, and these flags.

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
apetro: Good catch, adjusted in dfd58cf7999a8d8769eb8e3e771f7a8857743c17 .

PR #392, opened 2016-02-03
Abandoned : Search results tweaks

apetro:
Grab bag of tweaks to polish up search results.

# The tweaks

## Show loading indicator in place of results count when loading

It's kind of subtle, but here's the gif

loading indicator in search tab bar

## Suppress no-results search results sections on All tab

Show the search sections either when

1. They have results, or
2. They are the focused-upon search section.

That is, don't show them on the `all` tab when they have no results, since their presence in the search tabs is enough to let the user know what kinds of results had none.

suppress empty results on all tab

## Show the inline heading and `<hr>` only on the All tab

So, 

show heading on all tab

but

suppress heading

## Suggest browse as a way to recover from no-results MyUW search

suggest browsing to recover from no-results myuw search

## Only show the "Show all N directory results" control when there are more to see

So,

see all 17 results

but

directory see more

because the user can expand that right inline and see everything there is to see for that directory query:

directory see less

## Some style tweaks

End no-results statements with a period, since they feel rather, well, final.

Use sentence for the "See more" and "See less" controls.

directory see more

directory see less

# History of this changeset

There was #390 (merged) that polished up search results. There was #391 (closed un-merged) suggesting some tweaks to #390. This is a rebirth of #391 against `master` rather than against the now-merged feature branch behind #390.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
@@ -23,7 +23,7 @@
     <hr ng-show="activeFilter == 'all'">
     <loading-gif data-object='myuwResults'></loading-gif>
     <div ng-show="myuwFilteredResults.length == 0" class='no-result'>
-      No MyUW results
+      No MyUW results. <a href="/web/apps">Try browsing instead?</a>
     </div>
     <div ng-repeat="portlet in myuwFilteredResults = (myuwResults | filter:searchTermFilter | showApplicable:showAll | orderBy:sortParameter | limitTo:searchResultLimit)" class="result">
       <h4><a href="{{::portlet.maxUrl}}" target="{{::portlet.target}}">{{ portlet.name }}</a></h4>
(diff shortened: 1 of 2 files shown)


apetro:
(Rebased against `master`).

vertein:
There are some directory result things that have come up.  I'm going to post a new directory result only PR with some fixes.  I will probably grab some of the commits that are in here that relate to wisc directory results.  Will update the pr with a comment when that happens.

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
```diff
@@ -5,23 +5,28 @@
...
+        <i ng-show="googleSearchLoading" class="fa fa-spinner fa-pulse"></i></a>
       </li>
     </ul>
   </div>
 
-  <div id="myuw-results" class='search-results-container'>
-    <h4 class="header">MyUW</h4>
-    <hr>
+  <div id="myuw-results" class='search-results-container'
+    ng-show="myuwFilteredResults.length > 0
```
vertein: I'm uneasy about this.  I think we may have talked about this always being there.  I'm open to being wrong on that.

apetro:
Closing to indicate not intending to update this PR to resolve the merge conflicts, reconcile what of this has been harvested elsewhere, etc.  The point of the PR was to communicate some ideas. Sometimes code is a way to communicate precisely. Seems to have worked here.

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
apetro: That's a great point. MyUW results may be a special case.

PR #397, opened 2016-02-03
Removes see all directory is already showing all

vertein:
Removes the see all # directory results if the number of results shown is equal or less than the limit.

So, you don't see the 'see all link' for [0,3]


apetro:
:+1:

jhanstra:
:+1:

PR #398, opened 2016-02-03
Loading indicators in search result tabs.

apetro:
Adds loading indicators to search result tabs.

Before:

before

After:

loading indicators on search tabs

(Screencasts from a throttled connection.)

I like that it helps users understand what search results they're waiting for, clearly differentiating zero results from dunno-how-many-results-yet, allowing them to decide whether to keep waiting or to start investigating what's come back, for edge cases where search results return slowly.

(History: this is another shot from an idea in #392 , #391. Note comment at https://github.com/UW-Madison-DoIT/angularjs-portal/pull/391#discussion_r51722530 . )


timlevett:
My only fear is its a lot of loading. Loading up top, down low, etc... Maybe just hide the count during loading?

PR #399, opened 2016-02-05
Marketplace Search Revamp

jhanstra:
Got bored and started playing around with how marketplace looks, specifically the search bar and header. Here's what I came up. Discuss!

Before:
image

After:
image


apetro:
:+1:

vertein:
:+1:

timlevett:
:+1:

timlevett:
From @jhanstra he is giong to cleanup mobile first

jhanstra:
Cleanup on mobile is ready
image

timlevett:
I'm confused on where you search on the mobile version.

apetro:
I agree with @timlevett . I *like* the aesthetics of this design. It's clean. But that search bar doesn't look enough like a search bar -- I bet if we user tested that, confusion would be demonstrated. @jhanstra @keirserrie got ideas on how to make something almost as beautiful but feeling a bit more inviting of "this is where you type your search query"?

apetro:
Looking at diff, this changeset removes `angularjs-portal-home/src/main/resources/endpoint.properties.example`. Which doesn't sound like it has much to do with Marketplace revamping. :smile: Inadvertent change?

angularjs-portal-home/src/main/resources/endpoint.properties.example
```diff
@@ -1,51 +0,0 @@
...
-
-#subreddit feed example
-uwmadisonreddit.uri=https://www.reddit.com/r/uwmadison/hot.json
-uwmadisonreddit.username=
-uwmadisonreddit.password=
-uwmadisonreddit.attributes=
-
-wiscdirectory.uri=http://www.wisc.edu/directories/json
-
-wiscedusearch.uri=http://ajax.googleapis.com/ajax/services/search/web
```
timlevett: don't delete this file!

jhanstra: Shoot! Be careful with `git add -A`.

angularjs-portal-home/src/main/resources/endpoint.properties.example
```diff
@@ -48,4 +48,4 @@ uwmadisonreddit.attributes=
 
 wiscdirectory.uri=http://www.wisc.edu/directories/json
 
-wiscedusearch.uri=http://ajax.googleapis.com/ajax/services/search/web
+wiscedusearch.uri=http://ajax.googleapis.com/ajax/services/search/web
```
timlevett: We like a new line at the end of each file. otherwise gh gives use that yucky x arrow thingy. Red bad.

angularjs-portal-home/src/main/resources/endpoint.properties.example
timlevett: Easy to do. I found doing a `git status` after a `git add --all` can help identify when doing things.

jhanstra:
What do y'all think of this?

Desktop:
image

Mobile:
image

timlevett:
I like this a lot. :+1:

apetro:
:+1:

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace.html
```diff
@@ -3,67 +3,74 @@
...
-    <p>Categories:</p><a ng-repeat="category in categories" href="" class="category-links" ng-click="selectFilter('category',category)" ng-class="{true: 'selected-category'}[categoryToShow === category]">{{category}}</a>
-  </div>
+  <div class="marketplace-body">
+    <div class="mp-header">
+      <div class="browse-help">
+        <p class="hidden-xs">What are you looking for?</p>
+        <p class="visible-xs"><i class="fa fa-search"></i></p>
+      </div>
+      <div class="browse-input">
+        <input type="text" class="hidden-xs marketplace-search" placeholder="e.g. Benefits Information" ng-model="searchText" select-on-page-load aria-label="Search bar enter the app you are looking for">
```
apetro: Note that this is a Madison-ism coded into the `marketplace.html` template. Not a deal-breaker, but as we continue to encourage adoption of `angularjs-portal` beyond MyUW, need a strategy for localization more compelling than forking the HTML templates. Viable text externalization strategies?

This specific instance is interesting in that it might be worthwhile to externalize this into a set of example, so that sometimes it would say "e.g. Benefits Information" but sometimes it would say "e.g. Courses".

Which leads to the idea that maybe this should be "e.g. {randomly selected name of app from set of apps the viewing user can both browse and subscribe}".  

Which is scope creep in a lot of ways, but also solves the problem that the "Benefits Information" example is irrelevant to the non-employees in our user base.

timlevett: I think the randomly selected app could over-complicate the intention of a simple placeholder. Being generic like Courses is a more palatable approach. Someday it would be grant to start implementing ngTranslate (think property files in the form of json mixed with directives and filters).

keirserrie:
Huge improvement here. Think we could ship this with more traditional search bar treatment, would also like to see the standard tab treatment that we see on other pages, browse doesnt need its own tab styling.

Here's an extreme minimalist approach--

screenshot 2016-02-12 09 02 09


Strip down existing browse page to look like search results page, keep same filters at least for now, and if you want to search, use the bar at the top, which will give you app results

jhanstra:
Merging, will look at new style for tabs and some of the other ideas mentioned in separate issues.

PR #406, opened 2016-02-16
Making user temperature choices persistent - MUMMNG-1524

Doug-Reed:


angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -43,22 +43,37 @@ define(['angular'], function(angular){
     populateWidgetContent();
   }]);
 
-  app.controller('WeatherController', ['$scope', 'layoutService', function($scope, layoutService){
+  app.controller('WeatherController', ['$scope', 'layoutService', 'keyValueService', '$q', function($scope, layoutService, keyValueService, $q){
     $scope.weatherData = [];
     $scope.loading = false;
     $scope.showMetric = false;
     $scope.currentlyImperial = true;
+    $scope.fetchKey="\"userWeatherPreference\"";
```
timlevett: The escaped quotes are not needed here.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -69,6 +84,21 @@ define(['angular'], function(angular){
...
+    	if(userPreference==='C'){
+    		$scope.changeToC();
+    		$scope.showMetric=true;
+    	}else{
+    		$scope.changeToF();
+    	}
+
+    	var value = {};
+    	value.userWeatherPreference = userPreference;
+       	keyValueService.setValue($scope.fetchKey, value);
```
timlevett: So on initial setting that you do on line 75, wouldn't this then immediately save? We would want to avoid this. Init flag param?

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -69,6 +84,21 @@ define(['angular'], function(angular){
...
       }
     };
+    
+    $scope.changePref=function(userPreference){
+
+    	if(userPreference==='C'){
+    		$scope.changeToC();
+    		$scope.showMetric=true;
+    	}else{
+    		$scope.changeToF();
```
timlevett: You could add in a `$scope.showMetric = false;` here and remove the double call in the `ng-click`

timlevett:
Little late to the party ;) , but :+1:

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -43,22 +43,38 @@ define(['angular'], function(angular){
     populateWidgetContent();
   }]);
 
-  app.controller('WeatherController', ['$scope', 'layoutService', function($scope, layoutService){
+  app.controller('WeatherController', ['$scope', 'layoutService', 'keyValueService', '$q', function($scope, layoutService, keyValueService, $q){
     $scope.weatherData = [];
     $scope.loading = false;
     $scope.showMetric = false;
     $scope.currentlyImperial = true;
+    $scope.fetchKey="userWeatherPreference";
```
apetro: Pro tip: Style: use whitespace consistently.  Cf. [Google JavaScript Style Guide](https://google.github.io/styleguide/javascriptguide.xml), noting that the most interesting part of it is not [its specific guidance about this formatting](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting), but 

> Parting Words

> BE CONSISTENT.

> If you're editing code, take a few minutes to look at the code around you and determine its style. If they use spaces around all their arithmetic operators, you should too. If their comments have little boxes of hash marks around them, make your comments have little boxes of hash marks around them too.

> The point of having style guidelines is to have a common vocabulary of coding so people can concentrate on what you're saying rather than on how you're saying it. We present global style rules here so people know the vocabulary, but local style is also important. If code you add to a file looks drastically different from the existing code around it, it throws readers out of their rhythm when they go to read it. Avoid this.

apetro:
:1: , this solution is looking great on `predev`.

PR #407, opened 2016-02-17
MUMUP-2361 ; Render on web

timlevett:
+ Added common function for marketplace to get the launchURL
+ Added launch url to marketplace and search
+ Added exclusive launch to widgets
+ cleaned up exclusive launch on compact
+ Added in mock interface to `/web/exclusive/feedback`
+ Added in mock entries.json that include `layoutObj` in the entry object

home page and search

marketplace


angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -166,6 +166,14 @@ define(['angular', 'jquery'], function(angular, $) {
                 }
             };
 
+            this.renderURL = function renderURL(portlet) {
```
apetro: :+1: . Love abstracting this into method.

apetro:
:+1: . I like this a lot. It's both the functional improvement that's going to allow incrementally rendering more portlets more better, and the code refactor that makes this easier to maintain.

I'm a little confused about the relationship of the `renderURL(..)` and `getLaunchURL(..)` methods. Opportunity to further refactor those into a single method, maybe up a level at the `Service` layer rather than in `Controller`s?

angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
timlevett: Its a step. Someday the marketplace and layout will be the same object, someday...

timlevett:
Thanks for the feedback @apetro . It is a bit odd, but the layout json object is different than the marketplace entry. Different variables even. So `getLaunchURL(portlet)` is for the layout page (used on /web/compact and /web/expanded) and `renderURL(marketplaceEntry)` is used by the marketplace/search pages.

I'd love to someday have those objects be the same, but out of scope for this pr. That would be a big undertaking.

PR #412, opened 2016-02-24
MUMMNG-2265 Make Email Surface clickable

Doug-Reed:


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/option-link.html
```diff
@@ -17,6 +16,5 @@
            <option ng-repeat='obj in portlet.widgetData[config.arrayName]' value="{{obj[config.value]}}" title='{{obj[config.display].length > 26 ? obj[config.display] : ""}}'>{{obj[config.display] | trimMiddle:26}}</option>
        </select>
     </div>
-  </div>
```
apetro: It's not obvious to me why this `<div>` is disappearing. That might be okay -- I can't understand everything. :) Or maybe that means something interesting is happening here worth articulating?

vertein:
:+1:

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/option-link.html
vertein: Talked f2f in 3233 because it was unclear to me too.  Looking at whole file made it clearer while talking f2f.

Doug-Reed: I experimented with where to put the href until I found the right spot - but I left the remains of my experiments scattered through the code while I was committing. Consequently, I committed some div tags I shouldn't have.

jhanstra:
:+1:

PR #413, opened 2016-02-25
cleanup from override change and feature change

timlevett:
+ Step 1 to using the `override.js`, changing `app-config.js` to have values (step 2 will be to pull out all angularjs-portal specific values into separate config and stop overwriting `app-config.js`)
+ Change `upBase` to `myuwHome` like what is in frame


apetro:
:+1:

PR #415, opened 2016-02-25
MUMUP-2395 : Change wording on launch/login for guest

timlevett:
+ Changed wording (based on feedback) to "Login".
+ Turns out that uPortal makes it so guest cannot add anything, so we only have the option of saying "you can launch the things that are on your home screen". I'd rather not play around with the render code :)
+ Also fixed margin so it aligned with the details button of the same styles

#### Before
http://goo.gl/rc7bcM

#### After
http://goo.gl/m4QIX7

_Note using mock data for after screenshot, so don't get too caught up with which ones you can launch vs. login_


vertein:
:+1:

jhanstra:
:+1:

keirserrie:
:+1:  works for me

PR #418, opened 2016-03-01
Configurable google search results

vertein:
Only shows google search if the user is in a configured group.  Otherwise no search result section shows. 

Example for River Falls user with no google search configured:
uwrfnosearchconfig

Example for River Falls user with google search configured:
uwrfgoogsearchconfig

There is some work to do so that the `No Matches` suggestions are configurable.


keirserrie:
:+1:

jhanstra:
:+1:

timlevett:
Super cool, :+1:

PR #419, opened 2016-03-02
Zero search result state

vertein:
Now when you get zero search results and you are not in the UW-Madison group, you'll only see the help that's configured.

Example of UWRF on search:
image

Example of UWRF on browse:
image

But of course, at UW we have the whole collection configured:
image

and
image


timlevett:
:+1:

Doug-Reed:
:+1:

keirserrie:
late to the party but :+1:

vertein:
Thanks @keirserrie !

PR #420, opened 2016-03-02
MUMUP-2407: Priority Notification

jhanstra:
Goes along with https://github.com/UW-Madison-DoIT/uw-frame/pull/173

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/list/partials/home-list-view.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/list/partials/home-list-view.html
@@ -1,4 +1,5 @@
 <div class="row portlet-frame">
+  <notification-bell mode="priority" class="visible-xs"></notification-bell>
   <home-header></home-header>
   <div role="main" class="col-xs-12 no-padding portlet-body" ng-controller="LayoutController as layoutCtrl">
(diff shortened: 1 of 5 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/list/partials/home-list-view.html
```diff
@@ -1,4 +1,5 @@
 <div class="row portlet-frame">
+  <notification-bell mode="priority" class="visible-xs"></notification-bell>
```
timlevett: Is there any way we can make this not be in the header of every single page? Maybe in the app-header or in the frame page?

apetro:
Observation: seems like `<notification-bell mode="priority" class="visible-xs"></notification-bell>` is being sprinkled into several partials. Opportunity to refactor to use fewer partials to implement the top part of the page, the neighborhood of the priority notification display?

jhanstra:
@apetro @timlevett it doesn't look like those partials were even necessary, since priority notifications are already attached to the header. I removed them. Now this changeset is really just supplying a few notifications as priority.

apetro:
:+1:

timlevett:
:+1:

PR #421, opened 2016-03-02
MUMUP-2439 Configure DoIT Help Desk URL as helpdeskURL for UW River Falls.

apetro:
( [MUMUP-2439](https://jira.doit.wisc.edu/jira/browse/MUMUP-2439) , tweak on changes in #419 .)

Per [KB documentation](https://kb.wisc.edu/page.php?id=15303), the DoIT Help Desk is the right contact for non-username-or-password-related issues for `my.wisconsin.edu` portal users, including UW River Falls.

By the time one sees the UW-River-Falls-specific search config, one is logged in, so the not-username-or-password-related hurdle is already jumped.

So configure the same `helpdeskURL` for UWRF as is configured for Madison.


vertein:
:+1: Suggest create a JIRA to help track this through the tiers as this release progresses.

apetro:
Good idea wrt tracking migration.  This tweaks #419 . There a JIRA tracking that this can piggyback, or this need a new JIRA (perhaps tracking both of these together?)

vertein:
If you change the commit message to add the JIRA key from https://github.com/UW-Madison-DoIT/angularjs-portal/pull/419 and include it in the predev commit, that will be fine.

PR #423, opened 2016-03-04
Stab at search.md for documenting search.

apetro:
I got to thinking about the need to document, understand options for adopting campuses that came up at sprint review yesterday.

Idea: documentation about `angularjs-portal` search features, in the `docs` directory.


timlevett:
:+1:

PR #425, opened 2016-03-10
MUMUP-2431 Trap exclusive mode error and display message.

Doug-Reed:
Server-side error was bypassing client-side failure function. Moved the $http.get(...) function to it's own method, and test the results of that.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/services.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/services.js
@@ -169,36 +169,27 @@ define(['angular', 'jquery'], function(angular, $) {
             );
         };
         
-        var attemptExclusiveCall = function(portlet, returnMe){
-            
+        
+        var getExclusiveMarkup = function(portlet) {
             return $http.get(SERVICE_LOC.context + '/p/' + portlet.fname + '/exclusive/render.uP',{ cache : true}).then(
                     function(result) {
-                        returnMe = result.data;
+                        var data = result.data;
                         if(data) {
                             portlet.exclusiveContent = $sce.trustAsHtml(data);
                             console.log(portlet.fname + "'s exclusive data came back with data");
+                        }else{
+                            portlet.exclusiveContent="<font color=\"red\">An error occured retrieving your data. This service is unavailable at this time.</font>";
                         }
+                        
                         return data;
                     },
                     function(reason) {
-                        return returnMe;
+                        miscService.redirectUser(reason.status, 'exclusive markup for ' + portlet.fname + " failed.");
                     }
                 );
         }
- 
-        var getExclusiveMarkup = function(portlet) {
-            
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/services.js
```diff
@@ -180,10 +181,22 @@ define(['angular', 'jquery'], function(angular, $) {
...
                     }
                 );
         }
+ 
+        var getExclusiveMarkup = function(portlet) {
+            
+            var returnMe=-1;
+            var data = attemptExclusiveCall(portlet, returnMe);
+            if(returnMe===-1){
+        	portlet.exclusiveContent="<font color=\"red\">An error occured retrieving your data. This service is unavailable at this time.</font>";
```
timlevett: @keirserrie @jjones770 - wanna take a gander at this?

angularjs-portal-home/src/main/webapp/my-app/layout/services.js
```diff
@@ -180,10 +181,22 @@ define(['angular', 'jquery'], function(angular, $) {
                         return data;
                     },
                     function(reason) {
-                        miscService.redirectUser(reason.status, 'exclusive markup for ' + portlet.fname + " failed.");
```
timlevett: We may still want to do the redirectUser function, that is in place in case the shib session is invalid so it redirects the user to the login page.

vertein:
A handy screenshot would help me visualize the error html

angularjs-portal-home/src/main/webapp/my-app/layout/services.js
```diff
@@ -180,10 +181,22 @@ define(['angular', 'jquery'], function(angular, $) {
...
+                        return returnMe;
                     }
                 );
         }
+ 
+        var getExclusiveMarkup = function(portlet) {
+            
+            var returnMe=-1;
+            var data = attemptExclusiveCall(portlet, returnMe);
```
timlevett: So we may have to chat about promises. Hit me up when you have a moment.

Doug-Reed:
This is how it looks with the dummy text. 

dummy error message

Doug-Reed:
Refactored the change to put it all back under one function, per @timlevett.

Doug-Reed:
revised error message

Doug-Reed:
@keirserrie Jess : Ok? See above.

Doug-Reed:
@timlevett : What do you want to do to the style?

timlevett:
Talked in chat, just reiterating:

http://getbootstrap.com/components/#alerts would be preferred for this.

`<div class="alert alert-danger" role="alert">...</div>`

Doug-Reed:
bootstrapped exclusive error

vertein:
:+1:

PR #426, opened 2016-03-11
MUMUP-2430 : Fix styling for widget card for static content

timlevett:
+ Renamed class for widget content
+ This also fixes an issue in exclusive mode where portlet-content is the other div in some portlet markup that is sent back.


apetro:
:+1:

vertein:
:+1:

PR #427, opened 2016-03-11
override.js is a thing

timlevett:
+ Switches `/web` over to using `override.js` instead of `app-config.js` as described in http://uw-madison-doit.github.io/uw-frame/v2.2.0/#/md/configuration
+ Since we had additional categories for config, I also created a `web-config.js` file that is just for angularjs-portal config, like search :)


timlevett:
Possible because of https://github.com/UW-Madison-DoIT/uw-frame/pull/187

PR #428, opened 2016-03-14
MUMMNG-1734 : Add widget creator to web

timlevett:
This is the first step in splitting out my-tools home into its own app. This also upgrades widget creator to use the latest widget code. They are one in the same now. So now we can do more than creating custom widgets with widget creator. We can demo some of the templates. Picked a couple high usage ones.

### Widget creator enhancements
+ Added a `widgetConfig` section for widgets with config
+ Added `widgetURL`, only used by RSS type for now
+ Added RSS, LOL, and Search with links types to widget creator

### Custom
http://goo.gl/ET9xA7
http://goo.gl/dXvqqS

### RSS
http://goo.gl/CCJ1OQ

### LOL
http://goo.gl/QoKJfD

### Search with Links
http://goo.gl/jd7ssk
http://goo.gl/4ji3L9

### DIFF
--- a/angularjs-portal-home/src/main/webapp/WEB-INF/web.xml
+++ b/angularjs-portal-home/src/main/webapp/WEB-INF/web.xml
@@ -2,32 +2,32 @@
  "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN"
  "http://java.sun.com/dtd/web-app_2_3.dtd" >
 
-<web-app 
+<web-app
     xmlns="http://java.sun.com/xml/ns/javaee"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
     version="2.5">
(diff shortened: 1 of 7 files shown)


keirserrie:
very cool :+1:

angularjs-portal-home/src/main/webapp/WEB-INF/web.xml
```diff
@@ -66,24 +66,24 @@
...
     <filter-mapping>
        <filter-name>ExpiresFilter</filter-name>
        <url-pattern>/*</url-pattern>
        <dispatcher>REQUEST</dispatcher>
     </filter-mapping>
-    
-    
+
+
     <!-- Keep this in sync with angular-page.js -->
```
apetro: What's `angular-page.js`, and does this changeset keep these `<servlet-mapping>`s in sync with it?

timlevett: Not sure, probably copied in long ago when this started.

timlevett: As you can see, i didn't add/modify that line during this change ;)

PR #431, opened 2016-03-18
MUMMNG-2404 Notify Portal Administrators that impersonate bugged.

apetro:
Use a Notification to notify Portal Administrators that user impersonation of Bucky profile is bugged.

Idea stems from @mrfarnham's understandable surprise that the feature was bugged, and thinking through how to effectively message about the issue in the context of its use.

`#dogfooding`. Effectively communicating this bug has some value, and there's also some value in the toy exercise of notifications.

If this idea goes forward, I'll add a sub-task to [MUMUP-2454](https://jira.doit.wisc.edu/jira/browse/MUMUP-2454)'s fixing of the bug to also expire out this notification.


apetro:
Skipping the within-product notification in favor of fixing the bug this sprint.

PR #434, opened 2016-03-24
Clarify that webPortletRender affects e.g. search results launch links.

apetro:
Adjust "/web portlet rendering" beta setting description to reflect that 

 * it's just expanded widgets that do not reflect `webPortletRender`, 
 * rather than just compact widgets that do reflect it.


timlevett:
Interesting it doesn't work from expanded, it should. Maybe we didn't do that, maybe that is bugged for some custom widgets. either way, :+1:

PR #436, opened 2016-03-25
Fix broken hyperlink to uw-frame docs.

apetro:


vertein:
:+1:

docs/markdown/home.md
```diff
@@ -1,6 +1,6 @@
 ### AngularJS-Portal Documentation
 
-Thanks for visiting the documentation for [angularjs-portal](https://github.com/UW-Madison-DoIT/angularjs-portal). This documentation describes the features and implementation details of the angularjs-portal project, the new home of MyUW. angularjs-portal is based on uw-frame, if you want to learn about that I suggest looking at [that docs page](madison-doit.github.io/uw-frame/).
+Thanks for visiting the documentation for [angularjs-portal](https://github.com/UW-Madison-DoIT/angularjs-portal). This documentation describes the features and implementation details of the angularjs-portal project, the new home of MyUW. angularjs-portal is based on uw-frame, if you want to learn about that I suggest looking at [that docs page][uw-frame docs].
```
timlevett: That isn't a valid URL either... should it be `http://uw-madison-doit.github.io/uw-frame/`

docs/markdown/home.md
```diff
@@ -22,3 +22,5 @@ Thanks for visiting the documentation for [angularjs-portal](https://github.com/
 #### Integration with uPortal
 + Overview
 + Customizations/Setup
+
+[uw-frame docs]: http://uw-madison-doit.github.io/uw-frame/
```
timlevett: Ohhh, i see what you did there. Odd syntax, confusing almost. May I suggest we just have it inline?

docs/markdown/home.md
apetro: This is a [reference-style link](https://daringfireball.net/projects/markdown/syntax#link).

You can [preview the resulting Markdown here](https://github.com/apetro/angularjs-portal/blob/fix-link-to-frame-doc/docs/markdown/home.md), I think the link works.

docs/markdown/home.md
apetro: You can. Done, in 428af89.

Choice of Markdown link syntax, definitely an implementation detail I don't need to have an opinion about.

I think I reflexively went for the reference syntax because I find it makes the prose read better. I use reference link syntax quite a bit in my own Markdown writing.

Now giving this way more thought than I ought to, though, I did err here in introducing the reference syntax -- the existing `[angularjs-portal](https://github.com/UW-Madison-DoIT/angularjs-portal)` link uses inline syntax, and of course an important style consideration on an edit is to fit with the style of the thing being edited.

PR #438, opened 2016-03-29
MUMUP-2450: View more web search results should open in new tab

jiayinjx:
Open with new tab instead of  launch on the current page,using `target=_blank`

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
@@ -120,7 +120,7 @@
 
     </div>
     <div>
-      <h4 ng-if="webSearchUrl && !googleEmptyResults"><a ng-href="{{webSearchUrl + searchText}}">View more results for {{searchText}} on {{domainResultsLabel}}</a></h4>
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
```diff
@@ -120,7 +120,7 @@ <h4 class='header'>{{domainResultsLabel}}</h4>
 
     </div>
     <div>
-      <h4 ng-if="webSearchUrl && !googleEmptyResults"><a ng-href="{{webSearchUrl + searchText}}">View more results for {{searchText}} on {{domainResultsLabel}}</a></h4>
+      <h4 ng-if="webSearchUrl && !googleEmptyResults"><a ng-href="{{webSearchUrl + searchText}}" target=_blank>View more results for {{searchText}} on {{domainResultsLabel}}</a></h4>
```
apetro: This will almost certainly work, but I'd have done ` target="_blank"` so that the attribute value is quoted in the same way that e.g. the `ng-href` attribute value is quoted. Consistency?

timlevett:
:+1: `target=_blank` sounds good

angularjs-portal-home/src/main/webapp/my-app/search/partials/search-results.html
timlevett: This would work (HTML5 makes quotes optional for attributes), but for consistency I agree, please quote.

timlevett:
Closing this in favor of #439

vertein:
#439 merged

PR #439, opened 2016-03-29
Open new tab

vertein:
So we can close https://github.com/UW-Madison-DoIT/angularjs-portal/pull/438 and get this merged now rather than later.


timlevett:
:+1:

PR #443, opened 2016-03-30
MUMUP-2478 turn of old my.wisc

vertein:
No longer overrides the back2classicURL and removes the footer url inviting users to go back to MtyUW old


timlevett:
:+1: seems legit

PR #444, opened 2016-03-30
Wiscard widget

vertein:
This configures local angular to work with new wiscard widget supplied here: https://github.com/UW-Madison-DoIT/uPortal/pull/471


timlevett:
:+1:

PR #445, opened 2016-03-31
App directory documentation.

apetro:
Document app directory entries.

See [preview of resulting documentation](https://github.com/apetro/angularjs-portal/blob/doc-app-directory/docs/markdown/app-directory.md).


timlevett:
:+1: May I suggest a squash before merging?

PR #449, opened 2016-04-14
MUMUP-2346 : See all link in outages rss widgets should open new tab

Daha5:
Adds the target attribute to be config.target, and then I will set the target to be equal to "_blank" in the outages entity files so that not all rss widgets are affected.


timlevett:
I just recalled that this config is also setup in the widget creator. Could you add it [here](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js#L380) as well so people creating new RSS widgets will know its an option?

Daha5:
Something like that?

timlevett:
:+1:

PR #450, opened 2016-04-21
Add in groups to features feed

timlevett:
Noticed we didn't upgrade the features feed for ajsp when we added the grouping feature.


vertein:
good catch :+1:

PR #456, opened 2016-04-26
Check if date is hidden, if so let the title go full width

timlevett:
before
http://goo.gl/vEEfcx

after
http://goo.gl/xKHx9G


vertein:
:+1:

PR #460, opened 2016-04-28
Actually true is false, and false is true

timlevett:
don't ask.


apetro:
I don't understand this change, but if I did, maybe I'd come to understand that `$scope.loading = true;` does not mean "It is true that the page is still loading." and rather means something else ("It's true that the page is as loaded as it's ever gonna be.", say.)

If `$scope.loading = true` does not mean "the page is truly loading" and `$scope.loading = false` does not mean "the page is not loading", well, is there some renaming available that would make this loading indicator code easier to understand?

timlevett:
Revamped `loading` to be `loaded`. Should make more sense now.

Doug-Reed:
:+1:

vertein:
👍

PR #463, opened 2016-05-12
MUMUP-2252 Add UWRF Directory Search

Doug-Reed:


Doug-Reed:
uwrfsearch

vertein:
We verified results when 0 state, error state, and too many state.

:+1:

PR #465, opened 2016-05-17
MUMUP-2548 : Google search

Doug-Reed:


vertein:
If you drop the callback param from the google search api, you don't get that comment at the beginning and then we don't need the whole cleaning function.

timlevett:
@vertein : but you need the callback if you are calling google directly due to cross site scripting (looking at the URL, going from `/web/api` to `https://...google...`

timlevett:
Another PR fixed this #466

PR #466, opened 2016-05-24
MUMUP-2548 Fixes wisc.edu google search

vertein:
Alternative to https://github.com/UW-Madison-DoIT/angularjs-portal/pull/465

Updates wisc.edu search to up-to-date api.  Differs in https://github.com/UW-Madison-DoIT/angularjs-portal/pull/465 by not using the callback param, so no need to parse out comment.

After:
image


timlevett:
:+1:

PR #467, opened 2016-05-26
Fixes the beta feature to filter things on your home screen

timlevett:
We had a beta feature we wanted to try out a while back with adding a filter to the home page. When we revamped the app menus we never fixed the code so that this beta feature worked. had a free couple minutes so thought I'd fix that.

#### Flag
http://goo.gl/fSpcLP

### See it in action
http://goo.gl/5b4MKN

### Desktop
http://goo.gl/SlhllV

### Mobile
http://goo.gl/xlDhdv


vertein:
:+1:

PR #468, opened 2016-06-01
MUMUP-2578 Source Wiscard widget screenshot from deployed /images .

apetro:
Applies the [staged-for-prod approach for sourcing this screenshot](https://git.doit.wisc.edu/myuw-overlay/angularjs-portal/merge_requests/20) to localhost, using the prod-deployed `/images` since `/images` is not reliably deployed to localhost.

Alternative solution considered but not adopted: sourcing the image directly from git.doit's web interface, figuring that localhost running against bleeding edge of that repo modulo git.doit availability would be good tradeoffs.


vertein:
directly hitting git.doit wouldn't be nice for that service.

I would rather just use the local image for localhost instead of hitting my.wisc.edu for the image.

timlevett:
Agree with @vertein

PR #469, opened 2016-06-01
If no launch text, don't show the button

timlevett:
For LOL, if they don't provide a launch text, it looks weird so lets just hide it.

### With Launch text
http://goo.gl/ypJ2K1

### Without Launch text
http://goo.gl/cDssrc


vertein:
:+1:

PR #471, opened 2016-06-01
Document lol widget launchText optionality

apetro:
In the documentation about the `list-of-links` widget type, note that `launchText` is now optional per #469. .


vertein:
:+1:

PR #472, opened 2016-06-01
MUMMNG-2547 Notify System users who are ineligible

vertein:
image

@keirserrie helped out with the text of the notification.

I'm really looking for the opinion of Will Crickman of linking to kb article : https://kb.wisc.edu/page.php?id=15303.  I'm reaching out to Will offline (I don't think he has a github account).


timlevett:
Seems legit :+1: I thought at one point we stopped doing these in the base project but i could be misinformed. I created a [retro card](https://trello.com/c/A17u6LD0) to chat about it then.

vertein:
Chatted with Will, we're good.

PR #473, opened 2016-06-13
replace uw specific notifications with demo data

timlevett:
\- UW Specific Notifications
\+ Demo Data

UW Specifics will be covered in overlays.


vertein:
🍔 👍

apetro:
Would there by upside in adding enough demo data to demonstrate the breadth of notifications features?

Say,

* dismissable, and not
* a priority notification, perhaps targeted to a no-members group so that the way one demonstrates this is by setting the ignore-group-memberships-in-filtering-notifications beta flag?

PR #474, opened 2016-06-21
Widget creator buffering

timlevett:
Clean up some of the onscreen help in widget creator based on feedback from @mrfarnham . Thanks Mike.


vertein:
:+1:

PR #479, opened 2016-07-08
MUMUP-2583 : Edge case where you don't have anything in your layout spins forever

timlevett:
This at least removes the loading gif when you get nothing back (just like in compact mode)


vertein:
:+1: X :100:

PR #482, opened 2016-07-13
Update details view per recommendations from Keir

thevoiceofzeke:
In this PR:
- Added angular material elements
- Layout uses angular material instead of boostrap
- Added a "start" and "prestart" script to the node package so we can run the app with just `npm start` (saves a couple keystrokes)
- Removed a bunch of "px" from "0px" in CSS
- Replaced hard-coded colors in `marketplace.less` with references to colors in `variables.less`

**To do**
- Remove unused CSS
- Adjust other marketplace views so they conform to the updates (ex. replace the "add to home" and "launch" buttons in the search results view with toned-down material versions)

### Screenshots

screen shot 2016-07-14 at 11 58 46 am

screen shot 2016-07-14 at 11 58 54 am

### DIFF
--- a/angularjs-portal-home/src/main/webapp/css/variables.less
+++ b/angularjs-portal-home/src/main/webapp/css/variables.less
@@ -214,7 +214,8 @@
 }
 
 
-
+/* Transitions */
+@button-transition: box-shadow .4s cubic-bezier(.25,.8,.25,1),background-color .4s cubic-bezier(.25,.8,.25,1);
 
 /* MyUW colors */
 @color1: #b70101;         // Badger Red
--- a/angularjs-portal-home/src/main/webapp/css/variables.less
+++ b/angularjs-portal-home/src/main/webapp/css/variables.less
@@ -22,9 +22,7 @@
    ========================================================================== */
 @white: #FFFFFF;
 @black: #000000;
-@primary: #c5050c;
-@primaryDarkened20: #920000;
-@accent: #066999;
+@categoryBlue: #066999;
 
 /*
  * Grayscale colors. 1 is lightest, 10 is darkest.
--- a/angularjs-portal-home/src/main/webapp/css/variables.less
+++ b/angularjs-portal-home/src/main/webapp/css/variables.less
@@ -22,6 +22,9 @@
    ========================================================================== */
 @white: #FFFFFF;
 @black: #000000;
+@primary: #c5050c;
+@primaryDarkened20: #920000;
+@accent: #066999;
 
 /*
  * Grayscale colors. 1 is lightest, 10 is darkest.
@@ -37,6 +40,7 @@
 @grayscale8: #696969;
 @grayscale9: #404040;
 @grayscale10: #333333;
+@grayscale11: #222222;
 
 /*
  * Primary color with 3 lightened shades and 3 darkened shades.
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
@@ -57,7 +57,8 @@
                             ( {{::portlet.userRated}} rating<span ng-if="portlet.userRated !== 1">s</span> )
                         </span>
                         <br>
-                        <md-button ng-hide="GuestMode" ng-click="openRating('sm',portlet.fname, portlet.title)" class="md-raised rate-button">Rate</md-button>
+                        <md-button ng-hide="GuestMode" ng-click="openRating('sm',portlet.fname, portlet.title)"
+                                   class="md-raised rate-button">Rate</md-button>
                     </div>
                     <div class="desc-item">
                         <h3 tabindex="0" class="md-title">Related Apps</h3>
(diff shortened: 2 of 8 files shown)


vertein:
Looks cool so far!

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -10,111 +11,90 @@
...
-    </div>
-  </div>
 
-	<div class="portlet-footer left">
-	 	<ul>
-	 		<li><a ng-href="{{backURL}}"><i class="fa fa-arrow-circle-o-left"></i> Back to {{backText}}</a></li>
-	 	</ul>
+	<div class="portlet-footer" layout="row" layout-align="start center">
+        <md-card-actions>
+            <md-button ng-href="{{backURL}}"><i class="fa fa-arrow-circle-o-left"></i> Back to {{backText}}</md-button>
```
apetro: Maybe an `aria-hidden="true"` on the `<i .. fa-arrow-circle-o-left...`? It's maybe not strictly speaking "decoration", but I don't think it adds enough meaning beyond the "Back to" text, so screen readers would still be best served hiding the icon?

timlevett:
Looks good so far. If you wanted to read more about the `app-header` checkout the docs http://uw-madison-doit.github.io/uw-frame/latest/#/md/directives . You may need to use the `app-header-two-way-bind` directive to get it to work. I also kind of like the no app-header for the details, so maybe run that by Jess Jones (Keir's replacement)

vertein:
testing out @jjones770 mentioning.  Everyone can ignore

angularjs-portal-home/src/main/webapp/css/variables.less
```diff
@@ -22,6 +22,9 @@
    ========================================================================== */
 @white: #FFFFFF;
 @black: #000000;
+@primary: #c5050c;
+@primaryDarkened20: #920000;
+@accent: #066999;
```
timlevett: Please keep in mind theme support for other schools when making these color choices. See css/theme directory.

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -1,120 +1,106 @@
...
-  </div>
+<div ng-controller="MarketplaceDetailsController as detailsCtrl" class="portlet-frame portlet-details-page" flex="100">
+    <md-card>
+        <div ng-if='loading' class="row details-header no-margin">
+            <loading-gif data-object='obj' data-empty='!loading'></loading-gif>
+        </div>
+        <div ng-if='error'>
+            <div class="alert alert-danger" role="alert" style='margin: 20px;'>
+                <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
+                <span class="sr-only">Error:</span>
```
apetro: ❤️  the attention to accessibility. This did make me go look at [ARIA guidance on alerts](https://www.w3.org/TR/wai-aria-practices/#alert) and [on labeling](https://www.w3.org/TR/wai-aria-practices/#LabeledBy). This "alternative access" approach of having one span for typical browsers and another span for screen readers, I don't think it's bad, might well be the best practice for all I know :) but would also like to  bear in mind approaches that make a shared element "responsive" to different ways of consumption.

angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -1,120 +1,106 @@
...
-    </div>
-  </div>
+        <!-- DETAILS HEADER -->
+        <md-card-header tabindex="0" layout="row" class="no-margin" ng-if='!loading && !error'>
+            <md-card-avatar layout="column" layout-align="center center">
+                <portlet-icon class="md-headline"></portlet-icon>
+            </md-card-avatar>
+            <md-card-header-text layout="column" layout-align="center start">
+                <span class="md-headline" hide-xs>{{::portlet.title}}</span>
+                <span class="md-title" hide-gt-xs>{{::portlet.title}}</span>
```
apetro: Perhaps tedious questions, certainly questions about things I am not expert in:

1. Is this a sign that the `md-headline` or `md-title` should itself be "responsive", behave differently at different screen widths? Seems like this is the same text in the same conceptual role that just wants to display differently depending on width.

2. This looks like an either/or construct. Would this be more literate?

```xml
<span class="md-title" show-xs>{{::portlet.title}}</span>
<span class="md-headline" hide-xs>{{::portlet.title}}</span>
```

thevoiceofzeke: 1. I used md-healine and md-title just because they're existing CSS classes from Angular Material. It may be better to write our own class, but in the short term I felt it was better to work within the framework and not clutter up the LESS files (since I'm still wrapping my head around them).

2. You would think so, but that change would result in the former span being displayed all the time. I think you'd have to write it like this (but I'm not sure even this would work):

```html
<span class="md-title" hide-gt-xs>{{::portlet.title}}</span>
<span class="md-headline" hide show-gt-xs>{{::portlet.title}}</span>
```

PR #489, opened 2016-07-25
Move all portal api calls to /<portal context>/api

timlevett:
+ Move all portal API calls to use `/api` only as `/web` is a myuw specific thing and we still have `/api`


vertein:
👍

PR #491, opened 2016-07-26
List-of-links with single link should fall back to normal display

thevoiceofzeke:
In this PR:
- List-of-links widgets will display as default widgets if the following criteria are met:
    - `widgetType` is list-of-links
    - there is only one link in `widgetConfig.links`
    - `altMaxUrl` is true
    - the URL of that link is the same as the portlet's `url` attribute

### Screenshot
screen shot 2016-07-26 at 1 58 33 pm

**Note:** I tested this by changing the mock data so that the Manager Time and Approval widget meets all the criteria mentioned above. I did not push those changes.


vertein:
:+1:

PR #492, opened 2016-07-27
MUMNG-2638 Remove campus chat connections

Doug-Reed:


timlevett:
Looks like there are merge conflicts, but 👍

PR #494, opened 2016-07-29
Hotfix materialized widget display issues

thevoiceofzeke:
Widget body is now 100% height with a 36px bottom margin (the height of the "launch" button). This should ensure that all of the widget's content is displayed, but it's not a perfect fix. I think the underlying structure of widgets should be revisited long-term. I also plan on materializing the `<circle-button>` directive, which should help. 

### Screenshot
screen shot 2016-07-29 at 11 46 13 am


vertein:
Cool :+1:

PR #495, opened 2016-08-03
MUMUP-1536 link ratings API from details page, if beta setting.

apetro:
Annette asked me to grab the ratings details for a particular app.

In the interest of enabling fishing:

Adds a beta setting:

link ratings api

If (default) not set, details page presents ratings as before.

wiscard balance details

If the beta setting is set, the details page presentation of ratings is enhanced with a hyperlink

got hyperlink-

to the raw JSON API.

```JSON
{"ratings":
  [
    {"rating":5,
     "review":"Exactly what you need to know about your Wiscard and the most important direct links for managing it - which aren't easy to find otherwise."}
  ]
}
```

Note that authorization is applied on the JSON API itself, so while anyone could flip the beta setting to get the hyperlink, this has no security implications.

The raw JSON API is probably a "good enough" administrative user experience for accessing this data for now, though of course it will be tempting for a talented front end developer ( @thevoiceofzeke ?) to give this a more beautiful experience. :)

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/controllers.js
@@ -248,7 +248,9 @@ define(['angular', 'jquery'], function(angular, $) {
 
     app.controller('MarketplaceDetailsController', [
         '$controller', '$scope', '$routeParams', 'marketplaceService',
-        function($controller, $scope, $routeParams, marketplaceService) {
+        'SERVICE_LOC',
+        function($controller, $scope, $routeParams, marketplaceService,
+                 SERVICE_LOC) {
 
           $controller('marketplaceCommonFunctions', { $scope : $scope });
 
@@ -256,6 +258,8 @@ define(['angular', 'jquery'], function(angular, $) {
               currentCategory=category;
               currentPage='details';
           };
+          
+          $scope.ratingPrefix = SERVICE_LOC.base + SERVICE_LOC.marketplace.base;
 
           var figureOutBackStuff = function() {
             var fromInfo = marketplaceService.getFromInfo();
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
@@ -55,7 +55,7 @@
                         <span ng-if="$storage.linkRatingsApi">
                             <rating ng-model="portlet.rating" readonly="true" class="rating"></rating>
                             
-                            ( <a ng-href="/portal/api/marketplace/{{portlet.fname}}/ratings">{{::portlet.userRated}} rating<span ng-if="portlet.userRated !== 1">s</span></a> )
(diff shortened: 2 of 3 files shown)


angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
```diff
@@ -52,7 +52,13 @@ <h3 tabindex="0" class="md-title">Description</h3>
                     </div>
                     <div class="desc-item">
                         <h3 tabindex="0" class="md-title">Ratings</h3>
-                        <span>
+                        <span ng-if="$storage.linkRatingsApi">
+                            <rating ng-model="portlet.rating" readonly="true" class="rating"></rating>
+                            
+                            ( <a ng-href="/portal/api/marketplace/{{portlet.fname}}/ratings">{{::portlet.userRated}} rating<span ng-if="portlet.userRated !== 1">s</span></a> )
```
timlevett: Suggest changing `/portal/api/marketplace` to utilize [`SERVICE_LOC.marketplace`](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/js/override.js#L31) object and [`SERVICE_LOC.context`](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/js/override.js#L27). Thinking of those uPortal peeps.

apetro: I don't see how to lay hands on `SERVICE_LOC` from this template? Would the deeper solution be to make `ratingApiUrl` a property of `portlet` so it'd be `{portlet.ratingApiUrl}`?

timlevett: In the init of `MarketplaceDetailsController` you would inject the provider SERVICE_LOC. Then in the init function you could setup a scope variable for `ratingPrefix` or something like that.

apetro: Done.

Doug-Reed:
:+1:

angularjs-portal-home/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -248,14 +248,18 @@ define(['angular', 'jquery'], function(angular, $) {
...
+                 SERVICE_LOC) {
 
           $controller('marketplaceCommonFunctions', { $scope : $scope });
 
           $scope.specifyCategory = function(category) {
               currentCategory=category;
               currentPage='details';
           };
+          
+          $scope.ratingPrefix = SERVICE_LOC.base + SERVICE_LOC.marketplace.base;
```
timlevett: Can you move this down into the init function?

apetro: #500 would do this move.

timlevett:
I'll take a stab at creating a front end for this. Want to try out the dialog thing in angular material anyway.

apetro:
Awesome. I'm very supportive of this as a passion project, and especially supportive of trying out new angular material technologies. Hurray.

Since the audience for the feature is very tight and all administrators, I don't think a better UI than that raw JSON service is immediately required. So maybe only do this because you want to and it'll be fun. :)

PR #497, opened 2016-08-03
MUMUP-2632 Fix load more always showing

timlevett:
+ Load More now only shows if there are entries that are hidden
+ Load More count resets to 20 when you change your filter search item
+ Change [tabs to 2 spaces](https://www.youtube.com/watch?v=cowtgmZuai0), to see just changes [click me](https://github.com/UW-Madison-DoIT/angularjs-portal/pull/497/files?w=1)


vertein:
:+1:

PR #498, opened 2016-08-03
Fix display of add-favorites widget

thevoiceofzeke:
Changes to widget styles caused the "add to home" widget to display incorrectly. This is a fix for that specific widget.

screen shot 2016-08-03 at 3 35 05 pm


vertein:
:+1:

PR #503, opened 2016-08-08
Set default theme to first one and tab to 'Welcome' in master-overrides.js 

bjagg:
Change defaultTheme to use first theme instead of group API service, which does not exist in base uPortal. Also changed to default uPortal tab.


timlevett:
👍 The group API service doesn't exist in base uPortal? If not we should contribute that back.

timlevett:
@bjagg - Sure enough, that wasn't in base uPortal. Fixing that in https://github.com/Jasig/uPortal/issues/703

PR #504, opened 2016-08-08
Remove removed-widget toast and widget hover color

thevoiceofzeke:
In this PR:
- Toast message no longer appears upon removing a widget from home
- Hovering a compact widget no longer changes the widget's text/icon colors
- Fixed a bug where hovering some widgets did not change the cursor to a pointer (now all widgets behave the same on hover)
- Added a CSS transition to the add-favorites widget so it's background color changes smoothly instead of instantly


vertein:
:+1:

PR #513, opened 2016-08-22
MUMUP-2670 : Cleanup silent login step

timlevett:
Make the silent login step somewhat skip-able to fix localhost heavy development and fix master.


vertein:
👍

PR #514, opened 2016-08-23
Adjust marketplace to better utilize theme colors, fix widget title spacing

thevoiceofzeke:
**In this PR:**
- Removed hard-coded colors from marketplace tabs (so other themes don't get red tabs)
- Added ng-styles to non-material DOM elements that leverage the current material theme
    - *Note: These are inline styles, so I'm viewing it as a short-term fix :/*
- Adjusted button appearance/layout in marketplace
    - We get to /apps by clicking "add more to home," so it makes sense that "add to home" should be the primary call to action for each marketplace entry
    - Accent color is only used in raised buttons (so it is a background color instead of text color) -- this will ensure better color contrast for some themes)
- Adjusted text of the expand widgets toggle to say what it actually does as simply as possible ("expand widgets")
- Adjusted spacing of widget titles to avoid clipping when text wraps
- Formatted marketplace controllers so indentation is less maniacal

### Screenshots
screen shot 2016-08-23 at 9 55 13 am
screen shot 2016-08-23 at 10 11 12 am
screen shot 2016-08-23 at 11 01 09 am


apetro:
"Research Project Resource Guide" is technically one character over our [documented maximum title length](http://uw-madison-doit.github.io/uw-frame/latest/#/md/text-guidelines) -- but I'm not sure what one would do about that to comply with the limit? "Research Project Resources"? "Research Project Guide"? Maybe the title limit needs to be a little longer to embrace the reality of the long names of some things in our university world?

So I agree with the style tweak to avoid clipping, it's a good fix. Long thing titles continue to be a design challenge.

angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-header.html
```diff
@@ -3,8 +3,8 @@
     <md-switch class="md-accent md-hue-2 app-header-toggle" ng-model="expandedMode"
                aria-label="Toggle expanded widget layout" ng-change="toggleMode(expandedMode)"
                style="margin-top:0;margin-bottom:0;">
-      <span hide-sm hide-xs>Show expanded mode</span>
-      <span hide-gt-sm>Show Expanded</span>
+      <span hide-sm hide-xs>Expand widgets</span>
+      <span hide-gt-sm>Expand Widgets</span>
```
apetro: Inconsistent capitalization? [Prefer sentence-case](https://richardcooke.info/prefer-sentence-case-to-title-case/)?

angularjs-portal-home/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -2,13 +2,15 @@
...
-       '$localStorage','$rootScope', '$scope', '$routeParams', '$timeout', '$location',
-       function(googleCustomSearchService, miscSearchService, layoutService, marketplaceService, miscService,MISC_URLS, $sessionStorage,
-        $localStorage, $rootScope, $scope, $routeParams, $timeout, $location){
+  app.controller('marketplaceCommonFunctions', ['googleCustomSearchService', 'miscSearchService', 'layoutService', 'marketplaceService',
+    'miscService', 'MISC_URLS', '$sessionStorage', '$localStorage','$rootScope', '$scope', '$routeParams', '$timeout', '$location',
+    function(googleCustomSearchService, miscSearchService, layoutService, marketplaceService, miscService, MISC_URLS, $sessionStorage,
+             $localStorage, $rootScope, $scope, $routeParams, $timeout, $location) {
+
+      var currentThemePrimary = $sessionStorage.portal.theme.materialTheme.primary['500'];
+      $scope.primaryColorRgb = 'rgb('+ currentThemePrimary.value[0] + ',' + currentThemePrimary.value[1] + ',' + currentThemePrimary.value[2] + ')';
```
timlevett: This could create issues if we didn't create a custom palette for the current theme. One may suggest using `$mdColors` instead: https://material.angularjs.org/latest/api/service/$mdColors

thevoiceofzeke: I tried using $mdColors, but I couldn't get it to work without a prohibitively massive amount of work. I'm not concerned about the potential palette problem at the moment because every single theme has a custom palette. I could write a check to ensure there's a materialTheme defined (fall back on black if not). Would that be okay?

timlevett: Yeah, check with proper default sounds perfect

apetro: I don't have an opinion about what the "fail gracefully" solution should be. I do greatly appreciate the attention to failing gracefully.

angularjs-portal-home/src/main/webapp/my-app/layout/partials/home-header.html
thevoiceofzeke: @apetro: It's written both ways at the moment because everything in md-menus is capitalized. I'm going to write a PR for frame that makes items in md-menus sentence-case as well (that would be consistent with google's guidelines).

PR #516, opened 2016-08-24
Improve cross-browser consistency of widget appearance

thevoiceofzeke:
Noticed that IE still had the clipping issue, where some of the widget content was not visible. Made a small change that should fix it for now. I still think the overall structure of widgets should be overhauled in the future.

### Screenshots (IE, then Chrome)
screen shot 2016-08-24 at 2 11 47 pm
screen shot 2016-08-24 at 2 11 55 pm


timlevett:
:+1:

PR #518, opened 2016-08-30
Fix login if guest (which returns failed from /portal/Login)

timlevett:
Minor bug where portal returns "failed" as the login for the guest user.


vertein:
:+1:

PR #520, opened 2016-08-31
[WIP] Enable js compression for web

timlevett:
Using r.js' compression setup, I'm trying to compress angularjs-portal's javascript files.

WIP because currently there are issues with when it happens. package step is too late and prepare-package is too early. Seems this may not work for overlays so I thought I would post this WIP to see if anyone had any thoughts.


timlevett:
some major issues with this approach as our services are not ready for this. Closing for now.

PR #521, opened 2016-09-01
Fix toggle using alternate hue of accent color

thevoiceofzeke:
Noticed during sprint review that "expand widgets" toggle was using a darker hue of the accent color, which is hard to see for some themes. This PR changes it back to using the default hue of the accent color.


timlevett:
👍

PR #532, opened 2016-09-22
Fix doc appearance again (we really need local testing)

thevoiceofzeke:


timlevett:
Created https://jira.doit.wisc.edu/jira/browse/MUMUP-2742 to get local testing for ajsp-docs

PR #534, opened 2016-09-27
Create directives for types of search results

thevoiceofzeke:
**In this PR:**
- New directive for each type of search results (only serves to reduce copy+pasted code)
- Fixed a bug that made the "load more" button for MyUW results never appear 
    - Currently in prod, if you search a term that would return more than 20 results (like the letter "r"), you will only get 20 results and the "load more" button will not appear
- Made the marketplace "load more" button sentence case
- Minor style tweaks
- Got rid of some unused jQuery in the search controller

screen shot 2016-09-27 at 11 30 05 am

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
@@ -19,7 +19,6 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
                 $scope.filterMatches = [];
                 return;
             }
-
             $scope.filterMatches = marketplaceService.filterPortletsBySearchTerm($scope.portlets, newVal);
         });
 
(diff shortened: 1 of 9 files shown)


angularjs-portal-home/src/main/webapp/my-app/search/controllers.js
```diff
@@ -87,46 +86,9 @@ define(['angular', 'portal/search/controllers', 'my-app/marketplace/controllers'
...
-        } else if (filterName == 'google') {
-          $('#google-selector').addClass('active');
-          $('#wisc-edu-results').show();
-          $('#wisc-edu-results-header').hide();
-          $('#myuw-results').hide();
-          $('#wisc-directory-results').hide();
-          $('#wiscDirectorySeeMoreResults').hide();
-        }
-      };
-
```
timlevett: yay for removing this.

PR #539, opened 2016-10-03
Fix widget rendering problems

thevoiceofzeke:
**In this PR**:
- structural html change to fix display of simple/normal widgets across browsers
- added an aria-label to widgets that allows screen readers to read "[app title] widget" instead of "unknown"

**Notes:**
- I was unable to test in IE because my VM is having issues. IE 10 and later support the Firefox draft of flex, so the `-webkit` prefixes should be sufficient. Older versions may require additional prefixing. This should be tested.
- Adding [postcss/autoprefixer](https://github.com/postcss/autoprefixer) to our build process will provide the following benefits: 
    - Browser inconsistencies will become irrelevant and we should see fewer (if any) issues like this
    - Our .less code will be easier to read and maintain (we won't have to have a ton of extra lines of code for vendor prefixes) <-- [backlogged MUMMNG-2853](https://jira.doit.wisc.edu/jira/browse/MUMMNG-2853)

### Screenshots (chrome, firefox, safari)
screen shot 2016-10-03 at 11 11 08 am
screen shot 2016-10-03 at 11 11 23 am
screen shot 2016-10-03 at 11 11 43 am

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -84,26 +84,30 @@
       </div>
 
       <!-- For basic apps (not a widget, not simple content, no pithy content), show only an icon -->
-      <a tabindex="-1" ng-switch-when="NORMAL" ng-href="{{widgetCtrl.renderURL(portlet)}}" target="{{portlet.target}}" class="normal-widget">
-        <div class="widget-icon-container" layout="column" layout-align="center center">
-          <portlet-icon></portlet-icon>
-        </div>
+      <div ng-switch-when="NORMAL">
+        <a tabindex="-1" ng-href="{{widgetCtrl.renderURL(portlet)}}" target="{{portlet.target}}" class="normal-widget">
+          <div class="widget-icon-container" layout="column" layout-align="center center">
+            <portlet-icon></portlet-icon>
+          </div>
+        </a>
         <button aria-labelledby="goToApps-{{portlet.nodeId}} appTitle_portlet.title-{{portlet.nodeId}}" class="launch-app-button">
           <span ng-if='portlet.widgetConfig.launchText'>{{portlet.widgetConfig.launchText}}</span>
           <span ng-if='!portlet.widgetConfig.launchText'>Launch full app</span>
         </button>
-      </a>
+      </div>
 
       <!-- For simple content portlets, show only an icon -->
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -84,26 +84,30 @@
       </div>
 
       <!-- For basic apps (not a widget, not simple content, no pithy content), show only an icon -->
-      <a tabindex="-1" ng-switch-when="NORMAL" ng-href="{{widgetCtrl.renderURL(portlet)}}" target="{{portlet.target}}" class="normal-widget">
-        <div class="widget-icon-container" layout="column" layout-align="center center">
-          <portlet-icon></portlet-icon>
-        </div>
+      <div ng-switch-when="NORMAL">
+        <a tabindex="-1" ng-href="{{widgetCtrl.renderURL(portlet)}}" target="{{portlet.target}}" class="normal-widget">
```
apetro: What does this anchor look like to a screen reader? Only content is the icon?

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -84,26 +84,30 @@
...
-      </a>
+      </div>
 
       <!-- For simple content portlets, show only an icon -->
-      <a tabindex="-1" ng-switch-when="SIMPLE" ng-click="widgetCtrl.maxStaticPortlet(portlet)" class="normal-widget">
-        <div class="widget-icon-container" layout="column" layout-align="center center">
-          <portlet-icon></portlet-icon>
-        </div>
+      <div ng-switch-when="SIMPLE">
+        <a tabindex="-1" ng-click="widgetCtrl.maxStaticPortlet(portlet)" class="normal-widget">
```
apetro: Some accessibility concern as line `88`.

thevoiceofzeke: Should we backlog a separate story? I can make changes to this now, but it doesn't really apply to the work being done here.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
thevoiceofzeke: VoiceOver highlights the link and reads it as "Unknown." This is a problem.

vertein: Interesting...  Yeah, probably don't want screen readers or key navigation to hit something that's redundant.  At some point, we made the launch button optional.  I don't think that was an option for these normal widgets?

PR #548, opened 2016-10-10
Fix display of compact widgets

thevoiceofzeke:
Recent changes to the "widget info" tooltip didn't include compact widgets, and they broke. This PR fixes them.

#### Before
screen shot 2016-10-10 at 10 59 21 am

#### After
screen shot 2016-10-10 at 10 59 29 am

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/partials/default-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/partials/default-card.html
@@ -1,14 +1,17 @@
 <md-card class="list-content" id="portlet-id-{{::portlet.nodeId}}">
-  <div class='widget-info'>
-    <i title="Info" class="fa fa-info-circle"
-    tooltip="{{::portlet.description}}"
-    tooltip-trigger="mouseenter"
-    tooltip-placement="top"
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/partials/default-card.html
```diff
@@ -1,14 +1,17 @@
...
-    <i title="Info" class="fa fa-info-circle"
-    tooltip="{{::portlet.description}}"
-    tooltip-trigger="mouseenter"
-    tooltip-placement="top"
-    tooltip-popup-delay="200"></i>
-  </div>
-  <div class='widget-remove' ng-hide='GuestMode'>
-    <i title="Remove" class="fa fa-times portlet-options" ng-click="layoutCtrl.removePortlet(portlet.nodeId, portlet.title)"><a aria-label="Remove this app symbol" href="#"></a></i>
-  </div>
+  <md-button class="widget-action widget-info md-icon-button" aria-label="see a brief description of this app">
```
apetro: Better `aria-label` if it was clearer about what app would be described?

Alternatively, maybe the widget `<a` should be `aria-describedby` this tooltip?

PR #549, opened 2016-10-17
Fix broken launch buttons

thevoiceofzeke:
**In this PR**:
- "NORMAL" widgets now use the `<launch-button>` directive and have an `href` attribute
- "SIMPLE" widgets still use plain `<button>`, but they now have an `ng-click` attribute

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -87,12 +87,12 @@
 
       <!-- For basic apps (not a widget, not simple content, no pithy content), show only an icon -->
       <div ng-switch-when="NORMAL">
-        <a tabindex="-1" ng-href="{{widgetCtrl.renderURL(portlet)}}" target="{{portlet.target}}" class="normal-widget">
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -87,15 +87,16 @@
...
           </div>
         </a>
-        <button aria-labelledby="goToApps-{{portlet.nodeId}} appTitle_portlet.title-{{portlet.nodeId}}" class="launch-app-button">
-          <span ng-if='portlet.widgetConfig.launchText'>{{portlet.widgetConfig.launchText}}</span>
-          <span ng-if='!portlet.widgetConfig.launchText'>Launch full app</span>
-        </button>
+        <launch-button data-href="{{ widgetCtrl.renderURL(portlet) }}"
+                       data-target="_blank"
+                       data-button-text="{{ portlet.widgetConfig.launchText ? portlet.widgetConfig.launchText : 'Launch full app' }}"
+                       data-aria-label="goToApps-{{portlet.nodeId}} appTitle_portlet.title-{{portlet.nodeId}}">
```
apetro: I think `data-aria-label` is [intended to be a human-ready label](http://uw-madison-doit.github.io/uw-frame/latest/#/md/directives), but it looks like identifiers here?

PR #551, opened 2016-10-20
Adjust styles to compensate for uw-frame CSS changes

thevoiceofzeke:
**In this PR:**
- Added CSS to compensate for styles that were removed from uw-frame in [#339](https://github.com/UW-Madison-DoIT/uw-frame/pull/339)


apetro:
Does anything bad happen if we merge this in advance of #339 ?

thevoiceofzeke:
@apetro: Nope, should be good to go.

PR #557, opened 2016-12-12
Improve portal home accessibility

thevoiceofzeke:
[MUMUP-2807](https://jira.doit.wisc.edu/jira/browse/MUMUP-2807)

**In this PR:**
- Added aria-labels to elements that were being skipped by screen reader
- Changed the "info" button's role to a tooltip and provide the tooltip text as the button's aria-label

Overall this results in a smoother experience when using a screen reader, with clearer context when navigating by keyboard.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -36,7 +36,7 @@
          </div>
          <div ng-bind-html="portlet.widgetContent"></div>
       </div>
-      <a class="launch-app-button" href="{{portlet.url}}" target="{{portlet.target}}">Launch full app</a>
+      <a class="launch-app-button" ng-href="{{ portlet.url }}" target="{{ portlet.target }}">Launch full app</a>
     </div>
 
     <div ng-switch="widgetCtrl.portletType(portlet)" class="widget-type-container">
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
@@ -3,9 +3,9 @@
   <!-- HEADER -->
   <md-card-header class="widget-header">
     <!-- Widget Chrome -->
-    <md-button class="widget-action widget-info md-icon-button" aria-label="see a brief description of this app">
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -107,7 +109,7 @@
             <portlet-icon></portlet-icon>
           </div>
         </a>
-        <button aria-labelledby="goToApps-{{portlet.nodeId}} appTitle_portlet.title-{{portlet.nodeId}}" class="launch-app-button"  ng-click="widgetCtrl.maxStaticPortlet(portlet)">
+        <button aria-label="{{ portlet.widgetConfig.launchText ? portlet.widgetConfig.launchText : 'Launch full app' }}" class="launch-app-button"  ng-click="widgetCtrl.maxStaticPortlet(portlet)">
```
apetro: I think the intent of the `aria-labelledby` fanciness was to associate the name of the specific app to be launched with the launch button even in the case where there's no special `launchText` configured. Working from that intent, should the default `aria-label` in the new version reflect the name of the specific app to be launched rather than defaulting to `'Launch full app'`?

thevoiceofzeke: I attempted to address this by referring to the widgetConfig, putting the responsibility on the widget creators, rather than on our widget templates. I think that's the best option from a separation-of-responsibilities perspective (i.e. If we're going to offer the ability to set aria-labels in the widgetConfig, we should encourage the use of that feature). 

I made my initial decision based on our [launch text guidance](http://uw-madison-doit.github.io/angularjs-portal/latest/#/md/widget-launch-button), but I think Jess would agree that a more meaningful fallback for aria-labels would be a good thing. Any objections to defaulting to `Launch {{ portlet.title }}`?

apetro: I think there (should be) different guidance for sighted vs screen-reader-facing text. 

For sighted, "Don't repeat yourself" is decently good advice - the app card's title already visually associates with the launch button, so the launch button should observe the Thumperian Principle and say as little as possible if it has nothing interesting to say.

Contrastingly, non-sighted users will have difficulty visually associating the card title with the launch button on that card -- at best one would have to remember what card context you're in between the two elements; at worst it'll just be an out-of-context generic "click here" style link. So for the `aria-label` being as specific as possible, even at the risk of repetition of what's already in the card title, is the way to go.

PR #571, opened 2017-02-02
Adjust docs Jekyll theme

thevoiceofzeke:
**In this PR**:
- switched to the `minimal` theme (cleaner slate, no custom fonts)
- added a custom default layout
- moved some content into a "sidebar"
- added some custom styles

### Screenshots

screen shot 2017-02-02 at 12 26 49 pm
screen shot 2017-02-02 at 12 27 05 pm


docs/Gemfile.lock
```diff
@@ -0,0 +1,205 @@
+GEM
```
apetro: If this file is to be `.gitignore`d (as it should be as I understand it), then this file shouldn't also be checked into the repository.

docs/_layouts/default.html
```diff
@@ -0,0 +1,55 @@
...
+  {% endif %}
+
+</header>
+<section>
+
+  {{ content }}
+
+</section>
+<footer>
+  {% include sidebar.html %}
```
apetro: Include `sidebar.html` in `<footer>`? This deliberate?

docs/_layouts/default.html
```diff
@@ -0,0 +1,55 @@
...
+  <link rel="stylesheet" href="{{ '/assets/css/style.css?v=' | append: site.github.build_revision | relative_url }}">
+  <meta name="viewport" content="width=device-width">
+  <!--[if lt IE 9]>
+  <script src="//html5shiv.googlecode.com/svn/trunk/html5.js"></script>
+  <![endif]-->
+</head>
+<body>
+<header>
+  <h1>
+    <a href="http://uw-madison-doit.github.io/angularjs-portal/" aria-label="go to docs home page">{{ site.title | default: site.github.repository_name }}</a>
```
apetro: Consider `href="/"`, which should amount to the same thing but Just Work ™️ if the repo moves or if someone copies out the files of this fine theme to implement GitHub Pages for another project.

docs/_layouts/default.html
thevoiceofzeke: Yes. I know it's not semantically pretty, but it's the easiest way to make the layout work for us without hacking the theme apart completely, because the minimal and dinky themes are float-based.

thevoiceofzeke: In retrospect, I could just rename the file "footer-aside.html" or something along those lines, for the sake of clarity.

Doug-Reed:
:+1:

PR #573, opened 2017-02-03
MUMUP-2832 Misc fixing of doc links

vertein:
- Fixes octocat image (did search for other images)
 - Fixes the return to home link


Doug-Reed:
:+1:

PR #577, opened 2017-02-16
AddingKelvin

Doug-Reed:
Weather widget now cycles between Fahrenheit, Celsius, and Kelvin. The user's last choice will persist.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
@@ -66,7 +66,6 @@ define(['angular'], function (angular) {
     $scope.fetchKey = "userWeatherPreference";
     $scope.initialPreference = true;
     $scope.currentUnits = 'F';
-    $scope.previousUnits = 'K';
     $scope.nextUnits = 'C';
 
     var populateWidgetContent = function () {
@@ -107,26 +106,26 @@ define(['angular'], function (angular) {
       var userPreference = $scope.nextUnits;
       if(userPreference === "" || userPreference === null || typeof userPreference === "undefined"){
         userPreference = 'F';
+         $scope.initialPreference = true;
+      }else{
+         $scope.initialPreference = false;
       }
 
-      if(userPreference === 'F'){
+      if(userPreference === 'F'  && !$scope.initialPreference){
         $scope.changeKToF();
         $scope.currentUnits = 'F';
-        $scope.previousUnits = 'K';
         $scope.nextUnits = 'C';
       }
 
       if(userPreference === 'C'){
         $scope.changeFToC();
         $scope.currentUnits = 'C';
-        $scope.previousUnits = 'F';
         $scope.nextUnits = 'K';
       }
 
       if(userPreference === 'K'){
         $scope.changeCToK();
         $scope.currentUnits = 'K';
-        $scope.previousUnits = 'C';
         $scope.nextUnits = 'F';
       }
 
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
```diff
@@ -1,10 +1,7 @@
 
 
 <div class="weather-dropdown">
-  <a class="weather-not-clicked" ng-show="showMetric" href=""
-    ng-click="showMetric=false; changePref('F')">Change to &degF</a>
-  <a class=" weather-not-clicked" ng-show="!showMetric" href=""
-    ng-click="showMetric=true; changePref('C')">Change to &degC</a>
+  <a class=" weather-not-clicked" href="" ng-click="cycleUnits()">Change units to &deg{{nextUnits}}</a>
```
apetro: This is cute, adding more unit type options while incurring now bigger a UI footprint, arguably no more complicated a UI footprint.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -103,47 +102,85 @@ define(['angular'], function (angular) {
...
 
-    $scope.changePref = function (userPreference) {
+    $scope.cycleUnits = function (){
 
-      if (userPreference === 'C') {
-        $scope.changeToC();
-        $scope.showMetric = true;
-      } else {
-        $scope.changeToF();
+      var userPreference = $scope.nextUnits;
```
apetro: I don't get it. In line 90 `$scope.nextUnits` was set to `C` regardless of what came back from the preferences web service, wasn't it? So isn't the rest of `populateWidgetContent()` going to always clobber the user's stored preference?

Doug-Reed: Lines 91 - 93, there's a loop to call cyaleUnits() until $scope.currentUnits and $scope.nextUnits into alignment with the user's preferences. 

The data comes in as Fahrenheit, so I'm just initializing the values to an order the code is expecting before I call that loop. If the user's preference is F, then we avoid that loop altogether, and I need that C in nextUnits to be populated so it will display.

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
```diff
@@ -26,53 +23,49 @@
...
         </div>
+      </div>
     </a></li>
   </ul>
 
 
   <p class="credit">
     Powered by <a href="http://www.worldweatheronline.com/"
-      target='_blank' rel='noopener noreferrer'>World Weather Online</a>
+                  target='_blank' rel='noopener noreferrer'>World Weather Online</a>
```
vertein: more whitespace?

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/weather.html
```diff
@@ -26,53 +23,49 @@
...
   </p>
 </div>
 <a class="btn btn-default launch-app-button" href="{{::portlet.url}}"
-  aria-label="go to Weather">Launch full app</a>
+   aria-label="go to Weather">Launch full app</a>
```
vertein: more whitespace?

angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
vertein: If myPref.userWeatherPreference is K, line 91 cyleUnits is called because pref = K and curentUnits = F and nextUnits = C

So then line 106 sets userPreference to C?

PR #578, opened 2017-02-17
Weather login

Doug-Reed:
Fixed initialization error. Defaults to Fahrenheit for new user

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
@@ -92,7 +92,7 @@ define(['angular'], function (angular) {
               userPreference = 'F';
             }
             
-            while(myPref.userWeatherPreference != $scope.currentUnits){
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -104,12 +109,7 @@ define(['angular'], function (angular) {
...
       var userPreference = $scope.nextUnits;
-      if(userPreference === "" || userPreference === null || typeof userPreference === "undefined"){
-        userPreference = 'F';
-         $scope.initialPreference = true;
-      }else{
-         $scope.initialPreference = false;
-      }
+
 
       if(userPreference === 'F'  && !$scope.initialPreference){
```
vertein: do we need the check of initial preference anymore?  Looks like it's always set to true now.

Doug-Reed: initialPreference removed.

PR #580, opened 2017-03-10
Refactor widget types

thevoiceofzeke:
**In this PR:**
- "Generic" widgets are now called "custom" widgets, because that's what they are (this change is backwards compatible)
- "Normal" widgets are now called "basic" widgets, because calling them normal is discriminatory towards other widgets (only impacts portal developers)
- "Simple" widget type no longer exists. Basic widgets will render static content portlets if the same conditions are met as before
- Refactored the way we render static/exclusive portlets to simplify the process
- Layout and widget controller files auto-formatted and heavily commented for posterity
- Renamed "single-widget-component" to "widget-preview-component" because that's what it is
- Renamed "lol" to "list-of-links" because [same reason]
- Got rid of unused pithy content widget type
- Cleaned up some unused code
- Updated widget type documentation (with screenshots!), including a more detailed description of how to create a basic widget
- Fixed hover color of launch button for widget preview on details page

*Note: This branch was tested in predev*

### Screenshots
screen shot 2017-03-10 at 3 10 09 pm

### Rendering a static portlet

Academic Resources is a "staticContent" portlet (a portlet with static content and without an altMaxUrl). Here you can see it functions the same as before, except the layout files have been simplified.

static-refactor


angularjs-portal-home/src/main/webapp/css/widget.less
```diff
@@ -140,7 +140,7 @@ swl {
 }
 
 /* GENERIC WIDGETS */
```
davidmsibley: Probably need to change the comment to reflect the name

docs/widgets.md
```diff
@@ -13,27 +11,89 @@ Widgets can:
...
 
-### Advantages of widget types over custom widgets
+## Basic widgets
+
+The barebones widget provides an app title, a large icon, and a launch button with configurable text. It's a simple link to an app or external URL.
+
+![basic widget](./img/basic-widget.png)
+
+### Sample entity file
+
```
davidmsibley: Possibly include the enveloping portlet-definition tag?

davidmsibley:
:+1:

angularjs-portal-home/src/main/webapp/my-app/layout/widget/partials/widget-card.html
```diff
@@ -116,9 +95,9 @@
         </launch-button>
       </div>
 
-      <!-- For simple content portlets, show only an icon -->
+      <!-- For simple content portlets, show only an icon (RARELY USED) -->
```
apetro: Could be worth saying why it's rarely used, which I think is because most simple content portlets have an external alternativeMaximizedUrl.

apetro: Could "SIMPLE" widgets be refactored to *be* "BASIC" widgets with the `widgetCtrl.renderURL(portlet)` call returning the `/web/widget/{fname}` path?

docs/widgets.md
apetro: Or link to a full-sized example file included with the docs?  Like, here's the most interesting snippet, and linked is the full example?

thevoiceofzeke: Considered including it, but Github markdown doesn't wrap code blocks, so it made the whole page really long. I'll go the link route.

vertein:
Good use of comments

PR #581, opened 2017-03-13
Document Apereo Incubation status

apetro:
Captures some status and next steps from today's Apereo Incubation initial conference call.


Doug-Reed:
:+1:

Issue #582, opened 2017-03-15
Static Analysis and Linting


ChristianMurphy:
Related to ["Citizens for a better uPortal tomorrow"](http://apetro.ghost.io/citizens-for-a-better-uportal-tomorrow-tomorrow/) and [UP-4745](https://issues.jasig.org/browse/UP-4745).

Static analysis and linters are a [great way to detect code smells and errors](https://www.microsoft.com/en-us/research/publication/static-analysis-tools-as-early-indicators-of-pre-release-defect-density/) before releasing code.

Apereo uPortal has been moving over to validating code with the Google code style, angularjs-uportal and uw-frame could also benefit from the lint checking goodness.

Most of the languages used in this project have linters.
If there is interest I could put together a Pull Request to setup some static analysis tooling.

---

For Java [Checkstyle](http://checkstyle.sourceforge.net/) configured with [Google code style](http://checkstyle.sourceforge.net/google_style.html) is a reliable linter.

---

For JavaScript [ESLint](http://eslint.org/) is a great option, it has a [Google code style configuration](https://github.com/google/eslint-config-google), as well as plugins to better check specific use cases and libraries ([Angular](https://github.com/Gillespie59/eslint-plugin-angular), [Jasmine](https://www.npmjs.com/package/eslint-plugin-jasmine), [RequireJS](https://www.npmjs.com/package/eslint-plugin-requirejs), [Promise](https://github.com/xjamundx/eslint-plugin-promise), [Browser Compatibility](https://github.com/amilajack/eslint-plugin-compat), [JavaScript in Markdown](https://github.com/eslint/eslint-plugin-markdown), and [JSON](https://github.com/azeemba/eslint-plugin-json))

---

For Less checking [Stylelint](https://stylelint.io/) with a [standard config (based off google css style)](https://github.com/stylelint/stylelint-config-standard) has been fairly reliable.

---

For Markdown [remark lint](https://github.com/wooorm/remark-lint) with [recommended](https://github.com/wooorm/remark-lint/tree/master/packages/remark-preset-lint-recommended) and [consistent](https://github.com/wooorm/remark-lint/tree/master/packages/remark-preset-lint-consistent) configuration can produce good results.

---

For Angular HTML templates [HTMLHint](https://github.com/yaniswang/HTMLHint) can have okay-ish results, [angular templates are difficult to lint properly](https://stackoverflow.com/questions/28150293/how-to-validate-angular-html-templates).

[...]

PR #586, opened 2017-03-16
Issue #582: Add ESLint to package.json test script

ChristianMurphy:
'posttest' automatically runs after 'npm test'

PR #587, opened 2017-03-16
Issue #582: Use 'eslint --fix' to apply linter recommendations

ChristianMurphy:
:notebook: Note that the Google styleguide recommends some ES6 features be preferred over their ES5 counterparts.
If you do not want to drop ES5 support or add a transpiler [the ES6 rules](https://github.com/google/eslint-config-google/blob/master/index.js#L279-L315) should be disabled. :no_entry_sign:


thevoiceofzeke:
We can't drop support for IE. Changes were requested in #586. We can review another `--fix` with es6 rules disabled.

PR #588, opened 2017-03-16
Issue #582: Add remark-lint to package.json test script

ChristianMurphy:


package.json
```diff
@@ -5,8 +5,9 @@
   "scripts": {
     "pretest": "mvn clean package",
     "test": "karma start angularjs-portal-home/target/web/karma.conf.js  --single-run",
-  	"prestart": "mvn clean package",
-  	"start": "mvn jetty:run"
+    "posttest": "remark .",
```
thevoiceofzeke: Should include `--ignore-path .gitignore`

thevoiceofzeke: Also would prefer if it was its own script (i.e. "lint-md")

package.json
```diff
@@ -35,6 +36,18 @@
...
+    "remark-cli": "^3.0.0",
+    "remark-lint": "^6.0.0",
+    "remark-preset-lint-consistent": "^2.0.0",
+    "remark-preset-lint-recommended": "^2.0.0",
+    "remark-validate-links": "^6.0.0",
     "requirejs": "^2.1.18"
+  },
+  "remarkConfig": {
+    "plugins": [
+      "preset-lint-recommended",
```
thevoiceofzeke: Default value for list-item-indent rule is creating a lot of irrelevant problems. Consider adding the following to the "plugins" array:

```json
[
    "remark-lint-list-item-indent",
    "mixed"
],
```

package.json
ChristianMurphy: @thevoiceofzeke can do!
Should `posttest` become `lint-js` to match the other tasks?

ChristianMurphy: @thevoiceofzeke ignore setting has been configured, `lint-md` task has been added, eslint has been moved to `lint-js`.

package.json
ChristianMurphy: @thevoiceofzeke Thanks!
That has been added.

PR #589, opened 2017-03-16
Issue #582: Add stylelint to package.json test script

ChristianMurphy:


ChristianMurphy:
@thevoiceofzeke Updated with lint tasks broken out of `test` step

PR #595, opened 2017-03-17
Issue #582: Use 'eslint --fix' to apply linter recommendations

ChristianMurphy:
same as #587, but without ES2015+ changes applied, to maintain IE support.

### DIFF
--- a/angularjs-portal-home/src/main/webapp/js/master-override.js
+++ b/angularjs-portal-home/src/main/webapp/js/master-override.js
@@ -1,83 +1,81 @@
 define(['angular'], function(angular) {
-
-  /*Keep in sync with docs/markdown/configuration.md*/
+  /* Keep in sync with docs/markdown/configuration.md*/
 
     var config = angular.module('override', []);
     config
-        //see configuration.md for howto
+        // see configuration.md for howto
         .constant('OVERRIDE', {
-          'APP_FLAGS' : {
+          'APP_FLAGS': {
             'enableToggle': true,
-            'defaultView' : 'expanded',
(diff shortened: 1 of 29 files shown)


angularjs-portal-home/src/main/webapp/js/master-override.js
```diff
@@ -1,83 +1,81 @@
...
-            'defaultTheme' : 0,
-            'debug' : true,
-            'showUserSettingsPage' : true
+            'defaultView': 'expanded',
+            'compact': true,
+            'expanded': true,
+            'isWeb': true,
+            'defaultTheme': 0,
+            'debug': true,
+            'showUserSettingsPage': true,
```
ChristianMurphy: @apetro @thevoiceofzeke I'm not sure how far back you support IE.
Dangling commas will work fine in IE8+, it will cause a parser error in IE7 and below.

thevoiceofzeke: @ChristianMurphy: Will discuss with team and get back to you. My hunch is it's probably okay to drop IE7 and earlier. Personally, dangling commas trigger me :P -- but I can learn to live with them.

thevoiceofzeke: Yep, our official position is we do not support anything earlier than IE10 :)

ChristianMurphy: @thevoiceofzeke I feel the same way about dangling commas, after dealing with it for a while the irritation wears off a bit.

Issue #599, opened 2017-03-23
Fix Lint Errors For TravisCI


davidmsibley:
I'm creating This issue to notify any developers currently working on this repository that I'm working on a wide-reaching pull request to fix lint errors.  This will touch most of the controllers, directives and services within the repository, and cause merge conflicts with any concurrent work.

(Hopefully you're not in the middle of anything @ChristianMurphy!)

ChristianMurphy:
Thanks for the heads up @davidmsibley! 
No JavaScript changes in progress at the moment.

One small note on the lint configuration, I noticed a small correction that should happen.

```
"angular/di": "error" => "angular/di": ["error", "array"]
```

I can put together a PR if you want, or it could be incorporated as part of your planned changes.

davidmsibley:
I'm on it, thanks!

I've changed that to
`"angular/di": "warn"`
for now.

Speaking with @thevoiceofzeke, we think the array syntax is there only due to historical precedence, so we're not going to enforce array.  However, we also won't force a change to function syntax at this moment either.  We'll encourage new edits to use function syntax and at some point in the future change that rule back to error.

ChristianMurphy:
Array syntax also has benefits for additional minification (it allows minifiers to mangle names).

PR #601, opened 2017-03-23
Fix Lint Errors for TravisCI

davidmsibley:
Broad but hopefully thin changeset for fixing the Lint errors we should fix, and being warned about the ones that we can fix as we go.  The goal is to stay as close to Google's eslint standards as we can. Rule Overrides will be documented as line comments in the pull request.

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,13 +6,11 @@
     "lint-css": "stylelint --syntax less angularjs-portal-home/src/**/*.less",
     "pretest": "mvn clean package",
     "test": "karma start angularjs-portal-home/target/web/karma.conf.js  --single-run",
-    "dev-lint-js": "eslint --ext js --ext json --ext md --format html --output-file angularjs-portal-home/target/web/lint-out.html . --ignore-path .gitignore",
     "lint-js": "eslint --ext js --ext json --ext md . --ignore-path .gitignore",
     "lint-md": "remark . --ignore-path .gitignore",
     "lint-less": "stylelint --syntax less angularjs-portal-home/src/**/*.less",
     "prestart": "mvn clean package",
-    "start": "mvn jetty:run",
-    "watch-lint": "live-server angularjs-portal-home/target/web/lint-out.html"
+    "start": "mvn jetty:run"
   },
   "repository": {
     "type": "git",
@@ -29,7 +27,6 @@
   },
   "homepage": "https://github.com/UW-Madison-DoIT/angularjs-portal",
   "devDependencies": {
-    "concurrently": "^3.4.0",
     "eslint": "^3.17.1",
     "eslint-config-google": "^0.7.1",
     "eslint-plugin-angular": "^1.6.2",
@@ -50,7 +47,6 @@
     "karma-jasmine-html-reporter": "^0.2.2",
     "karma-phantomjs-launcher": "^1.0.4",
     "karma-requirejs": "^1.1.0",
-    "live-server": "^1.2.0",
     "phantomjs": "^2.1.7",
     "remark-cli": "^3.0.0",
     "remark-lint": "^6.0.0",
--- a/package.json
+++ b/package.json
@@ -6,11 +6,13 @@
     "lint-css": "stylelint --syntax less angularjs-portal-home/src/**/*.less",
     "pretest": "mvn clean package",
     "test": "karma start angularjs-portal-home/target/web/karma.conf.js  --single-run",
+    "dev-lint-js": "eslint --ext js --ext json --ext md --format html --output-file angularjs-portal-home/target/web/lint-out.html . --ignore-path .gitignore",
     "lint-js": "eslint --ext js --ext json --ext md . --ignore-path .gitignore",
     "lint-md": "remark . --ignore-path .gitignore",
     "lint-less": "stylelint --syntax less angularjs-portal-home/src/**/*.less",
     "prestart": "mvn clean package",
-    "start": "mvn jetty:run"
+    "start": "mvn jetty:run",
+    "watch-lint": "live-server angularjs-portal-home/target/web/lint-out.html"
   },
   "repository": {
     "type": "git",
@@ -27,6 +29,7 @@
   },
   "homepage": "https://github.com/UW-Madison-DoIT/angularjs-portal",
   "devDependencies": {
+    "concurrently": "^3.4.0",
     "eslint": "^3.17.1",
     "eslint-config-google": "^0.7.1",
     "eslint-plugin-angular": "^1.6.2",
@@ -47,6 +50,7 @@
     "karma-jasmine-html-reporter": "^0.2.2",
     "karma-phantomjs-launcher": "^1.0.4",
     "karma-requirejs": "^1.1.0",
+    "live-server": "^1.2.0",
     "phantomjs": "^2.1.7",
     "remark-cli": "^3.0.0",
     "remark-lint": "^6.0.0",
@@ -60,7 +64,7 @@
   "stylelint": {
     "extends": "stylelint-config-standard",
     "rules": {
-      "selector-type-no-unknown": false
+      "selector-type-no-unknown": null
     }
   },
   "remarkConfig": {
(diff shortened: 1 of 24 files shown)


package.json
```diff
@@ -102,24 +102,35 @@
     ],
     "rules": {
       "angular/angularelement": "error",
-      "angular/di": "error",
-      "angular/di-order": "error",
+      "angular/di": "warn",
```
davidmsibley: Google's standard is to use function syntax rather than array syntax for dependency injection.  However, IIRC function syntax was not always available, and I'm guessing that we kept using array syntax due to momentum.

We believe that eventually we'd like to switch to function syntax to comply with Google's standard.  We should backlog a discussion about this.  @ChristianMurphy mentioned that array syntax is actually useful for minifiers to be able to minify DI variables.

package.json
```diff
@@ -102,24 +102,35 @@
...
+      "angular/di-order": "warn",
```
davidmsibley: This is mostly stylistic, so it can stand to be a warning for now.  We intend to bite the bullet at some point and flip this back to error.

ChristianMurphy:
@davidmsibley Two helpful things I have found in past ESLint styleguide migrations, which can hopefully be helpful here:

1. having the linter integrated into the IDE helps. <http://eslint.org/docs/user-guide/integrations#editors>
2. having a script to be able to focus on a single rule at a time helps. <https://github.com/IanVS/eslint-nibble>

package.json
```diff
@@ -102,24 +102,35 @@
...
       "angular/di-unused": "error",
       "angular/document-service": "error",
       "angular/log": "error",
       "angular/module-dependency-order": "error",
       "angular/no-service-method": "error",
-      "angular/no-services": "error",
+      "angular/no-services": "warn",
```
davidmsibley: We ran into a legacy usage of $q.all() and determined it was outside the scope of this pull request to fix it. We should backlog that fix and flip this back to error as soon as it's done.

ChristianMurphy:
:notebook: resolves #599

package.json
```diff
@@ -102,24 +102,35 @@
...
       "angular/window-service": "error",
       "compat/compat": "error",
       "arrow-parens": "off",
       "constructor-super": "off",
       "generator-star-spacing": "off",
+      "max-len": [
+        "error",
+        {
+          "code": 120,
```
davidmsibley: This one's my own subjective opinion, I welcome a challenge to this.  I feel that a line length of 80 is far too short for programming today.  In recent years we've shifted to value descriptive function and variable names.  Combine that with the base level of indentation in files, and I believe this rule becomes unnecessarily restrictive to developers.  I feel that upping the line length by 40 to 120 gives developers room to breathe, but keeps files to a reasonable width.

package.json
```diff
@@ -102,24 +102,35 @@
...
+          "tabWidth": 2,
+          "ignoreUrls": true
+        }
+      ],
       "no-new-symbol": "off",
       "no-this-before-super": "off",
+      "no-invalid-this": "off",
```
davidmsibley: This is a rule in base eslint:recommended that causes some trouble when used with Angular. It's common practice to use `this` within the controller function.

package.json
```diff
@@ -102,24 +102,35 @@
...
       "no-var": "off",
       "prefer-rest-params": "off",
       "prefer-spread": "off",
+      "require-jsdoc": "error",
```
davidmsibley: This is standard, but I added it here to show the difference between just having JSDoc comments, and making sure JSDoc comments are perfect.

package.json
```diff
@@ -102,24 +102,35 @@
...
+      "valid-jsdoc": "warn",
```
davidmsibley: Having correct, well formatted docs is a goal of ours, but outside the scope of this pull request.  I wasn't confident in my ability to comment the code correctly.  We should work on cleaning up the comments bit by bit, and flip this back to error.

davidmsibley:
I can't seem to find the comment, but I got the email, @ChristianMurphy, about eslint tooling, and dang do I wish I knew about https://github.com/IanVS/eslint-nibble before this! I had a watch running eslint every two minutes, and a live-server continually refreshing the results of the run.  Kinda hacky but it worked for this.

I had linter running in Atom though, which made fixing the individual errors much easier.

ChristianMurphy:
@davidmsibley I deleted the comment after I realized all the work was already done.
I posted a bit too late for this PR, sorry about that.

package.json
ChristianMurphy: personally I prefer to keep the `rules` section set only with rules that are being customized, but thats just me. :man_shrugging:

package.json
ChristianMurphy: @davidmsibley If complete and accurate JSDocs are a goal, [`eslint-plugin-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc) can offer some additional suggestions past what ESLint offers by default.

package.json
ChristianMurphy: Interesting I wasn't aware you were using the `this` controller convention.
the [`controller-as`](https://github.com/Gillespie59/eslint-plugin-angular/blob/master/docs/controller-as.md) rule in `eslint-plugin-angular` might be a good addition.

davidmsibley: Maybe I jumped the gun on that. @thevoiceofzeke any thoughts on this? A quick ctrl-f seems to imply we assign scope variables in controllers far more often than we use `this`

package.json
apetro: I'd favor not making an exception around line length. descriptive function and variable names are good -- and refactoring to better abstractions when 80 characters feels restrictive is good too.

I agree it's a question of style. The meta-value in not making exceptions about style is high. The first rule of code-style-club is don't talk about code-style-club. [`-0.9`](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions).

apetro:
@ChristianMurphy wrt https://github.com/UW-Madison-DoIT/angularjs-portal/pull/601#issuecomment-288810452 , it's just as much that we (and I'll take responsibility, *I*) communicated too late about the intention to do this work as it is you posted your solution too late.

I think there's going to be a learning curve, a transition period in how tactical plans are created and communicated as we all learn to treat `AngularJS-portal` more like a community open source product and less only as a MyUW team product. That's going to be a little messy and that's going to have to be okay. This PR, besides being an important step forward on its substantive merits, also is a meta-reminder of the opportunities in continuing to evolve how AngularJS-portal is developed.

package.json
davidmsibley: good enough for me. This was one of the first rules I changed to drop the number of errors I had to contend with.  I'm going to drop it back down to 80 locally and see if fixing that is feasible now that I fixed everything else.

Doug-Reed: I'm in favor of the expanded line width. Given the size of modern monitors, and the fact that we don't code on punch cards anymore - my .02 is that an 80 character limit feels archaic.

davidmsibley: I'm willing to put in a little extra effort to adhere closely to the Google spec now.  If we feel it's a hindrance in the future, we'll be better informed to make that decision then.

package.json
thevoiceofzeke: It's a mixed bag throughout the project. My personal preference is to use controllerAs and the `var vm = this`, avoiding `$scope` and `this` entirely (something I picked up from the [johnpapa Angular style guide](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md). This project is a long, long way from that though. 

In general I'm in favor of avoiding `this` inside angular controllers.

ChristianMurphy: @davidmsibley @thevoiceofzeke @apetro 
If you are targeting [johnpapa's styleguide](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md) I would recommend removing all the custom angular configuration I put in place and instead add [`eslint-config-angular`](https://github.com/dustinspecker/eslint-config-angular).

davidmsibley: I'd be up for that, but possibly as its own effort.  I dropped that in to my local and it looks like we're about 350 errors away from johnpapa's ideal.

PR #611, opened 2017-04-11
Add jsdoc for filteredArray widget function

thevoiceofzeke:
Found something that uses this function while editing entity files -- figured I should explain why it's in this controller in the first place!


angularjs-portal-home/src/main/webapp/my-app/layout/widget/controllers.js
```diff
@@ -363,8 +363,13 @@ define(['angular'], function(angular) {
...
     };
 
-    // Unused??
-    // TODO: Determine whether this is used or not
+    /**
+     * Filter array for provided values of a given object -- used by some widgets (i.e. Leave Balances)
+     * @param array The array to filter
+     * @param objectVar The array entry to search through
+     * @param strings The string values to test against each entry
+     * @returns {*} An array containing only the desired
```
ChristianMurphy: Maybe jsdoc type annotations could help?

``` javascript
/**
 * @param {Array<Object>} array The array to filter
 * @param {Object} objectVar The array entry to search through
 * @param {Array<String>} strings The string values to test against each entry
 * @returns {Array<Object>} An array containing only the desired
 */

PR #634, opened 2017-05-23
Detect and support layout.json v4-3

apetro:
Bit of progress peeled out from #633 .

Detects when the configured layout web service is uPortal DLM layout.json v4-3.

This implementation is ugly. It could maybe benefit specifically from a refactor to use `map` effectively, and definitely more generally from attention from someone more comfortable with idiomatic JavaScript.

More importantly: could benefit from refactoring to make the canonical representation of layout simply a flat array of String fnames, since that's all modern `uPortal-home` needs. That'd be huge.

Finally, no attempt at unit test coverage for this change. :/

----

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/services.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/services.js
@@ -110,26 +110,19 @@ define(['angular', 'jquery'], function(angular, $) {
               return result;
             });
             if (tabs && 0 < tabs.length) {
-
               var columns = tabs[0].content;
-
               var portlets = [];
-
               function collectPortletsFromColumns(column) {
                 portlets = portlets.concat(column.content);
               }
-
               if ($.isArray(columns)) {
                 columns.forEach( collectPortletsFromColumns );
               }
-
               result.layout = portlets;
-
             }
           } else if ($.isArray(data.layout)) { // layoutDoc
             result.layout = data.layout;
           }
-
           return result;
         };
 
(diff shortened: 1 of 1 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/services.js
```diff
@@ -96,6 +99,27 @@ define(['angular', 'jquery'], function(angular, $) {
             if (folders && 0 < folders.length) {
               result.layout = folders[0].portlets;
             }
+          } else if ($.isArray(data.layout.navigation.tabs)) { // v4-3
```
davidmsibley: Won't this throw an error if we're using layoutDoc, since data.layout.navigation is undefined?

davidmsibley:
`reduce` would probably be a better choice than `map` 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

But really I don't think there's a great need to refactor it even if it's "ugly" as long as it's explicit.

angularjs-portal-home/src/main/webapp/my-app/layout/services.js
```diff
@@ -96,6 +99,27 @@ define(['angular', 'jquery'], function(angular, $) {
...
+              var result = false;
+              if (el && SERVICE_LOC.layoutTab === el.name) {
+                result = true;
+              }
+              return result;
+            });
+            if (tabs && 0 < tabs.length) {
+              var columns = tabs[0].content;
+              var portlets = [];
+              function collectPortletsFromColumns(column) {
```
davidmsibley: just add a var declaration to this and the linter will shut up:
`var collectPortletsFromColumns = function collectPortletsFromColumns(column) {`

vertein: Good call!

PR #653, opened 2017-06-29
Remove phantom js runner from Karma

ChristianMurphy:
Chrome is now primary test runner.
Also https://www.bithound.io/github/ChristianMurphy/angularjs-portal/master/dependencies/npm/phantomjs#security-advisories

----


davidmsibley:
What's the real security risk with phantomjs? Considering it is only used within the testing, and all code that calls it is our own, I'd think those risks are mitigated.

That said, I have no opinion over which launcher we use.

apetro:
Hurray for removing dependencies we don't need. Let's worry about the realities of phantomjs security if and when there's a need for it to come back in as a dependency. :)

ChristianMurphy:
@davidmsibley https://github.com/UW-Madison-DoIT/uw-frame/pull/462 gives more in-depth explanation for the switch. :scroll: 
For the security side, its less that I see it as a risk, and more that I'm tired of seeing angularjs-portal flagged as insecure in the dependency security reports I run. :stuck_out_tongue_winking_eye:

PR #654, opened 2017-07-05
reverting to layoutdoc

davidmsibley:
layout.json is full of weird quirks, one of which makes testing locally not work at all. Reverting to layoutDoc for local builds.

----


davidmsibley:
This is the one where if layout.json believes a portlet doesn't exist/is failing, it filters it from the layout. This causes you to not be able to log in locally, because the Quickstart Accounts widget does not appear.

PR #672, opened 2017-08-06
refactor(npm): Restructure remarklint setup

ChristianMurphy:
Allow remarklint preset rulesets to import the version of remarklint they depend on.
Allow remarklint to fail the build if a warning occurs.
Style shortcut links to include trailing `[]`.

----


ChristianMurphy:
@davidmsibley my understanding is `[reference]` and `\[no a reference]` are too similar, so it further differentiates with a trailing `[reference][]`.

wooorm:
`[reference]` could end up being a link if there’s a definition somewhere later (`[reference]: http://example.com`), but you probably didn’t intend that, so the thinking goes that it’s better to be explicit (`[reference][]`) if you meant it, or escape otherwise!

ChristianMurphy:
Thanks @wooorm!

PR #674, opened 2017-08-09
Update configuration to use new messages model

thevoiceofzeke:
Related to: [uw-frame#484](https://github.com/UW-Madison-DoIT/uw-frame/pull/484)
Related to: [angularjs-portal overlay updates](https://git.doit.wisc.edu/myuw-overlay/angularjs-portal/merge_requests)

**In this PR:**
- Update override paths to use new app-config
- Update static messages source

----


angularjs-portal-home/src/main/webapp/js/override.js
```diff
@@ -28,11 +25,17 @@ define(['angular'], function(angular) {
...
             },
-            'notificationsURL': '/web/staticFeeds/notifications.json',
+            'messagesURL': '/web/staticFeeds/messages.json',
             'groupURL': '/portal/api/groups',
             'kvURL': '/storage',
             'loginSilentURL': '/portal/Login?silent=true&profile=bucky',
           },
+          'MESSAGES': {
+            'announcementsEnabled': true,
+            'notificationsEnabled': true,
```
apetro: Do we need these flags, or could the enablement of announcements and notifications be implied by whether there are announcement-flavored and notification-flavored items in the messages feed? If there's no announcements in the feed, then they're not enabled. If there's no notifications in the feed, then they're not enabled.

angularjs-portal-home/src/main/webapp/js/override.js
```diff
@@ -28,11 +25,17 @@ define(['angular'], function(angular) {
...
+            'groupFiltering': true,
```
apetro: Do we need this flag, or could group filtering be implied by whether there are groups in the items in the messages feed? If groups are specified, filter to them. If they're not, don't worry about it.

thevoiceofzeke: ¯\_(ツ)_/¯

Valid questions, but I don't feel confident that the decision to remove the flags should be up to me. As it is, I've just repurposed what was already there.

ChristianMurphy:
@thevoiceofzeke CI failure is coming for RemarkLint.
Cherry-picking or applying change from 372b343840662e8688f2fa03fb0845990e1c6b9a should resolve.

PR #686, opened 2017-09-07
build(maven): configure maven release plugin commit prefix

### LINKED ISSUE #682: Maven release plugin conflicts with commit message hook
the commit messages generated by the maven release plugin, cause the conventional commit check on commit messages to fail.
In order to release currently, the hook needs to be removed, then the release happens, then the hook is readded.
This is both frustrating and causes unneeded churn.

Some possible solutions:
1. configure the maven release plugin to use commit messages that align with conventional commits
2. configure the maven release plugin to skip the commitlint check when committing
3. configure commit lint to accept the messages autogenerated by the maven release plugin

ChristianMurphy:
this avoids a conflict between the conventional commit message linter and the maven release plugin.

resolves #682

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,7 @@
         <version>2.5.1</version>
         <configuration>
           <autoVersionSubmodules>true</autoVersionSubmodules>
+          <scmCommentPrefix>chore(release): </scmCommentPrefix>
           <useReleaseProfile>false</useReleaseProfile>
           <releaseProfiles>release</releaseProfiles>
           <goals>deploy</goals>


davidmsibley:
can we have it as `chore(release): [maven-release-plugin]` or will that still break the commit hook?

ChristianMurphy:
It would still break, this time with the title length.

```log
chore(release): [maven-release-plugin] prepare release angularjs-portal-parent-6.6.0
```

is 84 characters log, header max length is expected to be <72 characters.
ref: https://github.com/marionebl/commitlint/blob/e6ef0721a230b90fc17955b33983c8bea2767d7f/%40commitlint/config-angular/index.js#L7

ChristianMurphy:
@davidmsibley 
I'm taking an educated guess that you want to preserve information that these commits were autogenerated.
Would setting scm [`username`](https://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html#username) for the release commits to `Maven Release Plugin` communicate that effectively?

davidmsibley:
I'm trying to figure out if my hesitation here is rational. I know that we do have more than a couple of internal Jenkins jobs that parse commit messages and look for that standard maven release formatted commit message. But those jobs don't run on these repos, so I kind of think it'll be fine. Anytime you're changing from a standard format to custom, it warrants a little thought because it could hurt you down the line. I'm reminded of a recent discovery by a coworker that he couldn't easily import years of server logs into a log summarizer/visualizer because on day one they'd decided to tweak the apache log format a bit to add more information (thus breaking the OOTB log parser).

There may be some irrational fear here as well. At the moment I place little to no value in commit message linting, and I despise commit hooks. So the idea that we change our build system to accommodate them is unpleasant. Maybe someday in the near future the value of commit message linting will become clear and I'll love it the same way I changed my opinion on Greenhopper the moment it started opening pull requests.

tl;dr: I've become a curmudgeon, and maybe when @vertein and @apetro get back in to the office tomorrow you'll get a better review :smile:

apetro:
How much longer are we going to keep using Maven to release a product that's got nothing to do with Java? The sooner we turn the corner, the sooner this is moot and we get to re-solve these problems with other releasing tools. :smile:

PR #688, opened 2017-09-12
fix(search): remove domain google search due to removal of the service

davidmsibley:
Google removed the custom search api that we were using, and there doesn't seem to be a straightforward replacement for us.  Removing the error experience for now.

----


davidmsibley:
I'll try to do this as a configuration change. Making a configuration change will get up to production quicker.

I just commented the code out because it still may have some value.  It's very close to the code we'd use if we migrated to the custom search JSON API, so having it around would be advantageous in that scenario.

PR #695, opened 2017-09-15
Remove "categories" from app dir UI

apetro:
Categories are confusing. Stop bothering users with them.

**Removes navigation tab within app directory.**
removes-the-categories-tab

**Removes categories chips from app directory entries as rendered en masse.**
removes-this-noise

**Removes categories section from app directory details page.**
removes-this-div

Categories are unfit for purpose because they are tied up with permissions. Gotta have permission on the category to see the category qua category, but permission on the category inherits as permission on the portlets within the category. In practice this is confusing, yields to misconfigurations, accidental permission grants, a mess.

Maybe someday no-permission-complications "tags" could be re-introduced as a feature. Maybe we'll never need it.

But for near term it has become clear that we want to only grant permissions at the individual `portlet-definition` level and stop trying to get permissions on categories right.

This changeset just removes categories superficially from the UI. However, this will unblock removing categories whenever we can on the back-end with no weird UI consequences at that time.

----

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
+++ b/angularjs-portal-home/src/main/webapp/my-app/marketplace/partials/marketplace-details.html
@@ -91,17 +91,6 @@
                             <li ng-repeat="related in portlet.relatedPortlets"><a href="apps/details/{{ ::related.fname }}">{{ ::related.title }}</a></li>
                         </ul>
                     </div>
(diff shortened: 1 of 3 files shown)


apetro:
Agree on the screenshot. Localhost mock is apparently broken ( #696 ). So I suppose I'll have to get a "real" development environment going to take screenshots...

Doug-Reed:
:+1:

PR #703, opened 2017-09-26
fix: disabling options rendering on non-home paths

davidmsibley:
many thanks to @Doug-Reed for finding the bug and @thevoiceofzeke for helping fix it.

----

### DIFF
--- a/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
+++ b/angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
@@ -399,12 +399,12 @@ define(['angular', 'jquery'], function(angular, $) {
   }])
   .controller('OptionsController', [
     '$scope', '$location', function($scope, $location) {
-      var expanded = 'expanded';
-      var compact = 'compact';
+      var expanded = '/expanded';
+      var compact = '/compact';
       $scope.renderMe = false;
       if (
-        $location.url().indexOf(expanded) > -1 ||
-        $location.url().indexOf(compact) > -1
+        $location.url().indexOf(expanded) == 0 ||
+        $location.url().indexOf(compact) == 0
       ) {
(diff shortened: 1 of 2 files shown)


angularjs-portal-home/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -396,5 +396,17 @@ define(['angular', 'jquery'], function(angular, $) {
...
       vm.init();
+  }])
+  .controller('OptionsController', [
+    '$scope', '$location', function($scope, $location) {
+      var expanded = 'expanded';
+      var compact = 'compact';
+      $scope.renderMe = false;
+      if (
+        $location.url().indexOf(expanded) > -1 ||
+        $location.url().indexOf(compact) > -1
```
apetro: Is this a temporary or permanent fix? If permanent, any worries that someday there will be a "Mayflower Compact" app and weirdly the expand/contract widget is going to start showing on its details page? heh.

thevoiceofzeke: I think "permanent" in the same sense that our current solution to toggling modes is "permanent." 

That said, it's probably worth looking for a better long-term solution for toggling expanded/compact modes, which would presumably change this implementation as well.

davidmsibley: yeah, we're just finding all the edges of customization that we made in uw-frame to support angularjs-portal. This is as permanent as we want it to be. This will be temporary as long as we make efforts to transform angularjs-portal into "just another frame app".

davidmsibley: actually, you're right @apetro, this would break if any entities have expanded or compact in their paths. this should really be a startsWith

PR #609, opened 2017-03-31
MUMUP-2886 adding name transitions and descriptions

davidmsibley:
Proposing project name changes for incubation

----

Contributor License Agreement adherence:

- [x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

[Apereo CLA roster]: http://licensing.apereo.org/completed-clas
[Contributor License Agreements]: https://www.apereo.org/licensing/agreements

PR #715, opened 2017-09-30
Use name `uPortal-home` consistently

apetro:
Obsessively, consistently brand as `uPortal-home`.

~~Two~~ commits:

+ one for everything except actual artifact identifiers ("documentation"). This mostly won't break anything, though I suppose Jasig's picky `NOTICE` file checker will want downstream `NOTICE` files to reflect the change the next time the picky notice checker plugin runs on those projects.
+ ~~one for actual Maven artifactIds. BREAKING CHANGE. Using a goofy internal capital letter in these may be a bad idea, or at least may be unjustified.~~

~~Leading with thoroughly adopting the branding `uPortal-home` because this new name is young and now is the moment if we're going to do this, but **I could totally see dropping the artifact identifier changing commit from this and just doing the rest of the renaming** so as to not break any work that's been done to use the new artifact IDs and so as to not have weird internal caps in actual artifact IDs. Maybe artifact IDs just really ought to be all-lowercase.~~

Leading with thoroughly adopting the branding `uPortal-home` because this new name is young and now is the moment if we're going to do this, but **not renaming artifact identifies** so as to not break any work that's been done to use the new artifact IDs and so as to not have weird internal caps in actual artifact IDs. Artifact IDs just really ought to be all-lowercase.

For that matter: I'm not actually hung up on exactly what the name formatting is. I'm just hung up on adopting some specific formatting consistently, so that we have a brand.  
+ All-lowercase `uportal-home` would  be fine, I think. It's got different tradeoffs, steps away from the "uPortal" of the brand, but then getting rid of weird internal caps has some advantages. 
+ `uPortal-Home` would be fine. 
+ But I think `uPortal-home` is [what we said we'd do](https://groups.google.com/a/apereo.org/d/msg/uportal-dev/BXOjNeuTC7E/RfFa2_NZFgAJ). ( #609 ), so leading with that on this PR.

----

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to `uPortal Home`
+# Contributing to `uPortal-home`
 
 In short:
 
@@ -15,7 +15,7 @@ The [Apereo Welcoming Policy][] [applies][code of conduct].
 
 ## Contributor License Agreements
 
-As an [incubating, aspiring Apereo product][uPortal Home website on incubating], `uPortal Home` requires contributors and contributions to comply with [Apereo inbound intellectual property licensing practices][].
+As an [incubating, aspiring Apereo product][uPortal-home website on incubating], `uPortal-home` requires contributors and contributions to comply with [Apereo inbound intellectual property licensing practices][].
 
 The short version:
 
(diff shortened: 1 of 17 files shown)


davidmsibley:
:-1: for the artifactId, artifactId's should be all lowercase: https://maven.apache.org/guides/mini/guide-naming-conventions.html

The NOTICE file checker won't break as long as the names in the NOTICE file are the same as the "name" field in the pom.xml

:+1: for the docs change. Consistency is good. We had multiple ways we were spelling angularjs-portal in the old docs so I used those places to test out different formats of the name. I think we can have two formats: one for documentation that is referring to the package, and one format referring to the product.
`uportal-home` should always be used when referring to the package, and we can use `uPortal-home` when referring to the product. After messing around with multiple different formats, I kind of like `uPortal Home` and `uPortal App-Framework` when referring to the products. But ultimately, consistency is king, and I _may_ have kept the docs inconsistent in the the hopes that someone else would make the product name decision :smile:

apetro:
(Removed the artifact identifier change commit.)

PR #736, opened 2017-10-28
refactor: clarify primary color bootstrapping

apetro:
Refactors the change from #735 .

+ Refactor a conditional to consider the true case of the conditional first, dropping a negation.
+ Add comments narrating the primary color boostrapping logic.

I think I know *what* this is doing (and hope the comments I've added reflect a true understanding.)

But I don't get it. Is it okay to use `$rootScope.THEMES[0]`'s primary color rather than fully bootstrap the theming so that the user's preferred or assigned-by-group theme's primary color applies? Is this going to result in a weird experience for some users, say those who prefer themes with significantly different primary colors than that of `THEMES[0]`?

----


vertein:
Yeah, I guess I don't fully understand what's going in here either.  I'll try to get some time today to figure this out.

ChristianMurphy:
Hey @apetro, it looks like a merge conflict has cropped up. Would you be able to resolve the conflict?

PR #751, opened 2017-11-29
test(setup): configure to only run tests in uPortal-Home not in framework

vertein:
Currently the setup is to run tests in framework as well as app, this change sets it up so that it only runs the app tests by replacing files provided by the overlay.  This will resolve the failing tests here - https://github.com/uPortal-Project/uportal-home/pull/750

----

### DIFF
--- a/web/src/main/webapp/karma.conf.js
+++ b/web/src/main/webapp/karma.conf.js
@@ -1,0 +1,80 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


vertein:
I did some research and thought the same thing about Karma's file inclusion/exclusion rules.  I tried for a little while to use the rules.  I tried it here and in the uPortal-App-Framework and was running into weird require.js rules?  I'm open to missing something though.

ChristianMurphy:
> I did some research and thought the same thing about Karma's file inclusion/exclusion rules. I tried for a little while to use the rules. I tried it here and in the uPortal-App-Framework and was running into weird require.js rules?

Got it, I was hoping there might be a trivial refactor,  but that is not the case.
LGTM as is :+1:

PR #795, opened 2018-03-15
feat: add keyboard controls for widget layout

thevoiceofzeke:
[MUMUP-3249](https://jira.doit.wisc.edu/jira/browse/MUMUP-3249): "As a keyboard user I would like to re-order my layout so that I can customize my MyUW experience to my needs."

**In this PR**:
- Replace incompatible jQuery UI drag and drop implementation with angular-drag-and-drop-lists
- Using arrow keys while focusing a widget will move it up or down in the layout order
- Add initial focus point with aria label in list of widgets
- Add controls for compact mode
- Remove ui bootstrap layout implementation and replace with CSS-grid (with flex box fallback for unsupported browsers)

**Related PR:** [uportal-app-framework#721](https://github.com/uPortal-Project/uportal-app-framework/pull/721)

### New drag and drop
draggable

### Keyboard controls
keyboard-controls

### New responsive layout
new-layout

----

### DIFF
--- a/web/src/main/webapp/my-app/layout/controllers.js
+++ b/web/src/main/webapp/my-app/layout/controllers.js
@@ -47,6 +47,114 @@ define(['angular', 'jquery'], function(angular, $) {
     function($localStorage, $log, $sessionStorage,
       $scope, $rootScope, layoutService) {
       var vm = this;
+
+      /**
+       * Log whenever a widget is moved
+       * @param eventType {String} Kind of interaction that moved the widget
+       * @param fname {String} The moved widget's fname
+       * @param dropIndex {Number} Index where the widget landed
+       * @param startIndex {Number} (optional) Index before moving the widget
+       * @returns {boolean}
+       */
+      $scope.logEvent = function(eventType, fname, dropIndex, startIndex) {
+        switch(eventType) {
+          case 'dragEnd':
+            $log.info('Dragged ' + fname + ' to index ' + dropIndex);
+            break;
+          case 'keyboardMove':
+            $log.info('Moved ' + fname + 'from index ' + startIndex +
+              ' to index ' + dropIndex);
+            break;
+          default:
+            return true;
+        }
+      };
+
+      /**
+       * Respond to arrow key-presses when focusing a movable list element
+       * @param widget {Object} The widget trying to move
+       * @param event {Object} The event object
+       * @returns {boolean} True if proper conditions aren't met
+       */
+      $scope.moveWithKeyboard = function(widget, event) {
+        // get index independent of ng-repeat to avoid filter bugs
+        var currentIndex = findLayoutIndex($scope.layout, 'nodeId', widget.nodeId);
+        var previousIndex = currentIndex - 1;
+        var nextIndex = currentIndex + 1;
+        var widgetElement = document.getElementById('focus-handle__' + widget.nodeId);
+
+        // left or up
+        if(event.which === 37 || event.which === 38) {
+          // stop element from losing focus
+          event.preventDefault();
+          // if currentIndex is already 0, do nothing
+          if (currentIndex === 0) {
+            return true;
+          } else {
+            // remove item from the list
+            $scope.layout.splice(currentIndex, 1);
+            // reinsert at new index
+            $scope.layout.splice(previousIndex, 0, widget);
+            // todo: refocus moved widget
+
+            // save new layout order
+            saveLayoutOrder(previousIndex, $scope.layout.length, widget.nodeId);
+            // log change
+            $scope.logEvent( 'keyboardMove', widget.fname, previousIndex, currentIndex);
+          }
+        }
+        // right or down
+        if(event.which === 39 || event.which === 40) {
+          // stop screen from scrolling
+          event.preventDefault();
+          // if currentIndex is end of the list, do nothing
+          if (currentIndex !== $scope.layout.length - 1) {
+            // remove item from the list
+            $scope.layout.splice(currentIndex, 1);
+            // reinsert at desired index
+            $scope.layout.splice(nextIndex, 0, widget);
+            // save new layout order
+            saveLayoutOrder(nextIndex, $scope.layout.length, widget.nodeId);
+            // log change
+            $scope.logEvent('keyboardMove', widget.fname, nextIndex, currentIndex);
+          }
+        }
+      };
+
+      /**
+       * After a widget is moved, save the new layout using
+       * the given information
+       * @param dropIndex {Number} Where the widget ended up
+       * @param length {Number} Length of the layout array
+       * @param nodeId {String} ID of the moved widget
+       */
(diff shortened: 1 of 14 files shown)


web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -47,11 +47,127 @@ define(['angular', 'jquery'], function(angular, $) {
...
+        }
+      };
+
+      /**
+       * Respond to arrow key-presses when focusing a movable list element
+       * @param widget {Object} The widget trying to move
+       * @param event {Object} The event object
+       * @return {boolean} True if proper conditions aren't met
+       */
+      $scope.moveWithKeyboard = function(widget, event) {
```
ChristianMurphy: Could something like fluid drag and drop work here? https://wiki.fluidproject.org/display/fluid/Drag+and+Drop+Design+Pattern

Or could this be contributed upstream so that everyone can enjoy an accessible experience?

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -47,11 +47,127 @@ define(['angular', 'jquery'], function(angular, $) {
...
+
+      /**
+       * Find an array object with the given attribute/value pair
+       * @param array {Array} The array to iterate on
+       * @param attribute {String} The name of the attribute
+       * @param value {String} The value to match on
+       * @return {number} Index of the desired item or -1
+       */
+      var findLayoutIndex = function(array, attribute, value) {
+        for (var i = 0; i < array.length; i+= 1) {
```
ChristianMurphy: https://lodash.com/docs/4.17.5#findIndex

https://www.npmjs.com/package/lodash.findindex

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -239,7 +329,7 @@ define(['angular', 'jquery'], function(angular, $) {
       /**
        * Sets href attribute for 'BASIC' type widget
        * @param portlet
-       * @returns {*}
```
ChristianMurphy: https://github.com/gajus/eslint-plugin-jsdoc recommends `returns`. Either `return` or `returns` works for me.
Either way we should add that to the lint rules and make it consistent across the app.

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -267,6 +357,120 @@ define(['angular', 'jquery'], function(angular, $) {
...
+        }
+      };
+
+      /**
+       * Respond to arrow key-presses when focusing a movable list element
+       * @param widget {Object} The widget trying to move
+       * @param event {Object} The event object
+       * @return {boolean} True if proper conditions aren't met
+       */
+      $scope.moveWithKeyboard = function(widget, event) {
```
ChristianMurphy: How is this different from `web/src/main/webapp/my-app/layout/controllers.js` line 79-130? are both needed?

web/src/main/webapp/my-app/layout/controllers.js
thevoiceofzeke: It's in there already and triggers a warning. The instances changed in this PR were fixed automatically by running eslint with the --fix flag.

web/src/main/webapp/my-app/layout/controllers.js
thevoiceofzeke: Not necessarily, no. It's duplicated because the two views use different controllers, and refactoring those controllers isn't in the scope of this story. It's something we should do at some point.

web/src/main/webapp/my-app/layout/controllers.js
thevoiceofzeke: Personally I don't agree that sortable widgets is something that should be in the app framework. I don't even think widgets should be in the app framework at all. We included them because on-campus developers using the framework to develop other apps asked us to, and in retrospect I would have fought against doing it. 

Hopefully there comes a day when the app framework becomes a collection of API-driven web components. Then this separation of concerns won't be an issue. 🤷‍♂️

apetro: https://github.com/marceljuenemann/angular-drag-and-drop-lists/issues/348

ChristianMurphy: > I don't agree that sortable widgets is something that should be in the app framework.

I agree :+1:
Hence the suggestion to contribute it upstream at Angular drag and drop lists.
Thanks @apetro for providing the link to the issue. :bow:

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -47,11 +47,127 @@ define(['angular', 'jquery'], function(angular, $) {
...
       $scope, $rootScope, layoutService) {
       var vm = this;
+
+      /**
+       * Log whenever a widget is moved
+       * @param eventType {String} Kind of interaction that moved the widget
+       * @param fname {String} The moved widget's fname
+       * @param dropIndex {Number} Index where the widget landed
+       * @param startIndex {Number} (optional) Index before moving the widget
+       * @return {boolean}
```
apetro: What are the semantics of the boolean returned? `true` means what? `false` means what?

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -47,11 +47,127 @@ define(['angular', 'jquery'], function(angular, $) {
...
+          default:
+            return true;
+        }
+      };
+
+      /**
+       * Respond to arrow key-presses when focusing a movable list element
+       * @param widget {Object} The widget trying to move
+       * @param event {Object} The event object
+       * @return {boolean} True if proper conditions aren't met
```
apetro: What's up with the `boolean` return? `true` means didn't do anything? Is this idiomatic? In some Java methods, for instance, `true` means had an effect and `false` means didn't have an effect, e.g. [Set.add()](https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#add(E)) .

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -267,6 +357,120 @@ define(['angular', 'jquery'], function(angular, $) {
...
+       */
+      $scope.moveWithKeyboard = function(widget, event) {
+        // get index independent of ng-repeat to avoid filter bugs
+        var currentIndex =
+          findLayoutIndex($scope.layout, 'nodeId', widget.nodeId);
+        var previousIndex = currentIndex - 1;
+        var nextIndex = currentIndex + 1;
+
+        // left or up
+        if (event.which === 37 || event.which === 38) {
```
apetro: Magic constants? Worth abstracting? Can they be referenced from somewhere?

web/src/main/webapp/my-app/layout/controllers.js
thevoiceofzeke: Agree it's semantically meaningless, just the result of hasty writing. This is something I can clean up!

PR #798, opened 2018-03-21
fix: fail safe when portlet lacks title

apetro:
In case where `portlet.title` undefined
don't consider that `portlet` a filter match on `title`
but also don't exception out the whole filtering attempt.

The app directory data should be good. This should never be a problem. But if the data is bad in this way, the consequence should be contained to the handling of that one bad record.

----


ChristianMurphy:
:bellhop_bell: @apetro friendly reminder that this is close to being merge-able.

PR #799, opened 2018-03-21
fix: add .vscode to .gitignore

apetro:
Exclude local Visual Studio Code configuration from source control.

----


apetro:
Yup. Agree would be preferable to switch to autogenerated `.gitignore`. Went for the quick superficial fix rather than a deeper fix in this PR.

PR #801, opened 2018-03-22
docs(icons): Prefer Material icons and deprecate Font Awesome icons

apetro:
Since we prefer Material Icons, be really clear how configure `portlet-definition`s to use them.

Since we're pursuing Material Design, **formally deprecate Font Awesome icon support** in the hope of encouraging natural movement to the preferred icon vocabulary. **(This may be controversial.)**

(I'm not sure I knew how to make use of Material icons as app directory icons before seeing a colleague do it locally -- documenting the magic name of the `parameter` not least so I myself will come across this documentation next time I'm bootstrapping the icon for an app directory entry.)

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,9 +39,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Dependency upgrades
 * Updated app-framework to 9.0.1
 
-### Removed
-
-
 ## [8.0.0][] - 2018-03-21
 
 ### Added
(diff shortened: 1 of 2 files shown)


apetro:
The deprecating is far more important to me than the actual removing at this juncture. Reality is in MyUW we have a bunch of existing legacy use of Font Awesome icons. It's not like we can turn that off quickly. If we can stop incurring new non-Material icon usages and start seriously considering whether we can roll forward onto Material icon usages when we we touch things, that's a win.

The maintenance surface of this product cannot grow infinitely. Gotta trim away low value features where we can.

PR #805, opened 2018-03-23
feat: prompt confirmation/undo when removing widgets from home

thevoiceofzeke:
[MUMUP-3095](https://jira.doit.wisc.edu/jira/browse/MUMUP-3095): "As a person who just deleted an item from my MyUW homepage, I'd like acknowledgement that I did this and the opportunity to undo it, so I can understand what I did and recover if it's not what I really want."

**In this PR:**
- Removing a widget can be undone via a toast message
    - Toast resolves automatically after 3 seconds
    - If user clicks "Undo" before resolution, the widget is added back to the layout
- If user deletes another widget before the first toast has resolved, it will automatically resolve it and show a new toast
- On large screens, toast appears in top right of application space
- On small screens, toast appears at bottom of screen (per Material specs)
- Removed unnecessary `LayoutController`. Expanded and compact modes now share the same layout controller (`WidgetController`).
- Removed CSS-grid layout in favor of more widely supported flex box
- Updated existing layout controller unit test to test widget controller instead
- Animate widget removal/addition/move

**A note about the custom toast message:** The toast attempts to follow pattern guidelines for [Material Notifications](https://material.io/guidelines/patterns/notifications.html), so it is a custom version of Angular Material's "Toast" directive. All the customizations were made in one commit [9f3b253](https://github.com/uPortal-Project/uportal-home/commit/9f3b2534e5a78cae164faf83d5d42e8fefc1b7f9) to make reverting this work easier, if necessary/desired.

### Demos
confirm-remove-lg
confirm-remove-sm

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   uPortal's identity impersonation and attribute swapping features (#802)
 
 ### Changed
-* Removing widget from home screen now prompts for confirmation (allowing UNDO action) before saving changes (#805)
+* Removing widget from home screen now displays a toast allowing UNDO action before saving changes (#805)
 
 ### Fixed
 * Fixed bug that sometimes caused the wrong widget to be removed from user's layout (#804)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ### Changed
-
+* Removing widget from home screen now prompts for confirmation (allowing UNDO action) before saving changes (#805)
 
 ### Fixed
-
+* Fixed bug that sometimes caused the wrong widget to be removed from user's layout (#804)
 
 ### Dependency upgrades
 
--- a/web/src/main/webapp/my-app/layout/controllers.js
+++ b/web/src/main/webapp/my-app/layout/controllers.js
@@ -41,38 +41,65 @@ define(['angular', 'jquery'], function(angular, $) {
                                 $mdToast, $sessionStorage) {
       var vm = this;
       /**
-       * Show toast message confirming widget removal
+       * Remove widget from visible layout, then trigger confirmation toast
+       * before persisting the layout change
        * @param fname
        */
-      vm.removeWidget = function(fname) {
+      vm.removeWidget = function(title, fname) {
+        // Filter for fname match in layout
+        var result = $filter('filter')($scope.$parent.layout, fname);
+        var index = $scope.$parent.layout.indexOf(result[0]);
 
+        // Remove from layout
+        $scope.$parent.layout.splice(index, 1);
+
+        // Trigger confirmation toast
+        confirmWidgetRemoval(title, fname);
       };
 
-      var confirmWidgetRemoval = function() {
-        // layoutService.removeFromHome(fname).success(function() {
-        //   // Filter for fname match in layout
-        //   var result = $filter('filter')($scope.$parent.layout, fname);
-        //   var index = $scope.$parent.layout.indexOf(result[0]);
-        //
-        //   // Remove from layout
-        //   $scope.$apply($scope.$parent.layout.splice(index, 1));
-        //
-        //   // Clear marketplace flag
-        //   if ($sessionStorage.marketplace != null) {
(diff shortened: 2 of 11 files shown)


web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -85,14 +108,18 @@ define(['angular', 'jquery'], function(angular, $) {
        * @param event {Object} The event object
        */
       $scope.moveWithKeyboard = function(widget, event) {
-        // get index independent of ng-repeat to avoid filter bugs
+        if ($scope.isToastDisplayed) {
+          return;
```
apetro: Worth firing a supplemental toast or something to tell the user why their move attempt is being ignored?

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -125,327 +152,131 @@ define(['angular', 'jquery'], function(angular, $) {
...
+      $scope.$on('REMOVE_WIDGET',
+        /**
+         * Listen for widget removal event, then show confirmation toast
+         * if one is not already showing.
+         * @param event {Object} The angularjs event object
+         * @param data {Object} Data about the widget being removed
+         */
+        function(event, data) {
+          // Make sure no toasts are already open
+          if (!$scope.isToastDisplayed) {
```
apetro: Is silently ignoring `REMOVE_WIDGET` event the right behavior in the case where that event fires with a toast already displayed? Or if that were to happen is the right thing to go ahead and stack up another toast? (Looks like such an event should never happen.)

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -125,327 +152,131 @@ define(['angular', 'jquery'], function(angular, $) {
...
-            // sake of backwards compatibility,
-            // but return what it really is: CUSTOM
-            return 'CUSTOM';
-          } else if ('custom' === portlet.widgetType) {
-            return 'CUSTOM';
-          } else {
-            return 'WIDGET';
-          }
+        if ($sessionStorage.portal.theme) {
+          // theme already in session, use primary color from it
```
apetro: Maybe this is using `-accent` rather than primary color from the theme?

web/src/main/webapp/my-app/layout/list/partials/home-list-view.html
```diff
@@ -30,12 +30,14 @@
...
     <ul dnd-list="layout"
         dnd-horizontal-list="true"
+        dnd-disable-if="isToastDisplayed"
         class="layout-list compact">
       <li class="dndPlaceholder no-padding widget-container"></li>
       <li class="no-padding widget-container"
           tabindex="0"
           aria-label="{{ widget.title }}"
           keep-focus
+          ng-class="{ 'hidden' : $index === hideWidgetIndex }"
```
apetro: Will non-visual-browser users be able to understand that the widget is hidden?

web/src/main/webapp/my-app/layout/partials/toast-widget-removal.html
```diff
@@ -0,0 +1,41 @@
...
+      <span class="toast__icon-container">
+        <widget-icon></widget-icon>
+      </span>
+    </div>
+    <div class="toast__actions" layout="row" layout-align="start center">
+      <md-button class="md-accent" ng-click="undoRemoveWidget($event)" aria-label="undo remove">
+        Undo
+      </md-button>
+      <md-button class="md-accent" ng-click="confirmRemoveWidget()" aria-label="confirm remove">
+        OK
```
apetro: I'm not sure I like that the user has to confirm the change. I love that the user has opportunity to un-do. Would there be a way to do this that defaults to honoring the user's apparent intent once expressed yet nonetheless gives the user an easy way to un-do it in case they were misunderstood?

That is, I'm not sure a user should have to click twice to remove an item so much as I'm sure that once a user does remove an item they should be one click away from un-doing this.

CHANGELOG.md
```diff
@@ -13,10 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   uPortal's identity impersonation and attribute swapping features (#802)
 
 ### Changed
-
+* Removing widget from home screen now prompts for confirmation (allowing UNDO action) before saving changes (#805)
```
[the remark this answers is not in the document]
thevoiceofzeke: That was prior to rework -- updated

PR #808, opened 2018-04-09
fix: tweak snackbars per material guidelines

thevoiceofzeke:
[MUMMNG-4155](https://jira.doit.wisc.edu/jira/browse/MUMMNG-4155): "As a user who recently deleted a widget from my layout, I would like the opportunity to undo my action before the snackbar disappears, so that I may undo a delete if I want."

[Material guidelines for snackbars & toasts](https://material.io/guidelines/components/snackbars-toasts.html#)

**In this PR**:
- Increased time before auto-hide
- Removed customizations that made snackbars look like notifications
- No longer show widget icons in snackbar message
- Snackbar always appears fixed to bottom of viewport (so it's always visible)

### Demo/screenshot
confirm-remove
screen shot 2018-04-09 at 12 24 32 pm

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [8.1.1][] - 2018-03-30
 
+
+### Changes
+* Tweaked "Undo" action toasts to better conform to []material guidelines](https://material.io/guidelines/components/snackbars-toasts.html#) (#808)
+
 ### Dependency upgrades
 * Updated app-framework to 9.0.2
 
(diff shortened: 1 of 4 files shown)


thevoiceofzeke:
I haven't found any specific guidance either, but both Angular Material projects use 3 seconds as the default delay before hiding. I think we have wiggle room to set it to whatever we feel is appropriate.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [8.1.1][] - 2018-03-30
 
+
+### Changes
+* Tweaked "Undo" action toasts to better conform to []material guidelines](https://material.io/guidelines/components/snackbars-toasts.html#) (#808)
```
apetro: typo in the hyperlink syntax around "material guidelines"

vertein:
I do like defaults.  We can see how this goes.  My gut says 4 seconds is probably enough, but we can probably see how this goes with actual users.

PR #815, opened 2018-05-08
WIP fix: removes rootScope.GuestMode

Doug-Reed:
removes rootScope.GuestMode in favor of a lighter weight way of passing that variable

----


web/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -57,10 +57,10 @@ define(['angular', 'jquery', 'require'], function(angular, $, require) {
           && portlet.portletName // there is a portletName
           && portlet.portletName.indexOf('cms') != -1; // static content portlet
       };
-
+      $scope.GuestMode = $sessionStorage.GuestMode;
```
davidmsibley: This might not be what we want. I think instead we'd like it to get `GuestMode` from the service, rather than sessionStorage. At the moment, I forget what that looks like in other places. I'll take a look this afternoon

davidmsibley: Yeah, we shouldn't replace this with a sessionStorage lookup, that's going to be removed as well.  You should be using the `mainService.isGuest()` async function. Here's an example: https://github.com/uPortal-Project/uportal-app-framework/pull/680/files#diff-5dc1012888a1a7946853bbacd81416b6R134

web/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -57,6 +57,16 @@ define(['angular', 'jquery', 'require'], function(angular, $, require) {
           && portlet.portletName // there is a portletName
           && portlet.portletName.indexOf('cms') != -1; // static content portlet
       };
+
+      var isGuest = function() {
```
ChristianMurphy: :art:

```js
var isGuest = mainService.isGuest;
```

ChristianMurphy: This defines a function `isGuest`, that leverages `mainService.isGuest`, and resolves the `result`.

web/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -58,9 +58,13 @@ define(['angular', 'jquery', 'require'], function(angular, $, require) {
...
       };
 
+      $scope.GuestMode = function(mainService) {
+        return mainService.isGuest;
+      };
+
       $scope.getLaunchURL = function(marketplaceEntry) {
         var layoutObj = marketplaceEntry.layoutObject;
-        if ($rootScope.GuestMode && !marketplaceEntry.hasInLayout) {
+        if ($scope.GuestMode && !marketplaceEntry.hasInLayout) {
```
ChristianMurphy: :question: Why is the check if `$scope.GuestMode` exists needed?

web/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -58,9 +58,13 @@ define(['angular', 'jquery', 'require'], function(angular, $, require) {
           && portlet.portletName.indexOf('cms') != -1; // static content portlet
       };
 
+      $scope.GuestMode = function(mainService) {
```
ChristianMurphy: :question: `GuestMode` is a factory that takes a service and returns a function?

web/src/main/webapp/my-app/marketplace/controllers.js
```diff
@@ -224,10 +228,18 @@ define(['angular', 'jquery', 'require'], function(angular, $, require) {
...
   .controller('MarketplaceController', [
     '$log', '$rootScope', '$scope', '$controller', 'marketplaceService',
-    function($log, $rootScope, $scope, $controller, marketplaceService) {
+    'mainService',
+    function($log, $rootScope, $scope, $controller, marketplaceService,
+      mainService) {
       var base = $controller('MarketplaceCommonFunctionsController',
         {$scope: $scope});
 
+      $scope.GuestMode = function(mainService) {
```
ChristianMurphy: :question: same as marketplace controller.js line 61?

PR #828, opened 2018-06-08
fix(search): show no-results div when no results

apetro:
Had been bugged such that no-search-results recovery suggestions would never show.
Change to show when there's no hope of results.

After: no results:

image 2018-06-08 at 7 37 09 am

After: some results

image 2018-06-08 at 7 39 07 am


apetro:
Glad you think so, @Doug-Reed ! I do think it'll help some. At the least maybe it will reduce user inhibitions to hit that "Give feedback" link and tell us if we're over-estimating how good this user experience is.

PR #829, opened 2018-06-08
fix(search): successfully launch into today.wisc.edu from Madison theme

apetro:
query parameter changed from `term` to `search[term]`


apetro:
Note automatic merge incorrectly merged `CHANGELOG` update such that this needs a manual fix commit.

PR #830, opened 2018-06-08
feat(search): deprecate Web search integration

apetro:
I think it's time to admit we're not going to fix and maintain this feature soon.

Removing it will reduce friction in other development and maintenance of the search experience.

And there's no rule we can't reach into source history and pick up and run with this feature whenever that becomes the right thing to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,14 +22,13 @@ and this project adheres to
 
 ### Deprecated
 
-**Web search integration is deprecated.** This integration is not actually in use
+**Web search integration is deprecated.** This integration is not actually used
 by `uPortal-home` adopters actively involved in maintaining this open source
-software product, so there's a lot of risk for this feature to go un-loved and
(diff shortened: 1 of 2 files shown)


davidmsibley:
charge us money, but also keep us on the old version of the search api. I'm sure that old version won't live forever, and in the mean time we would be losing out on the newer features. Our next move is to refactor to support the new API.

Doug-Reed:
There's a googleCustomSearchService in the marketplace controller in uportal-home. 

Can that come out as well?

apetro:
wrt to your scope of deprecation question, @Doug-Reed , In the most immediate next step, I'd say the dynamic, JSON-API-integrating aspects of this feature deprecate and drop out, whereas the just-paint-a-hyperlink-launching-external-web-search feature should remain, as that feature could work just fine if we'd put the link on the page where a user could click it.

But in the most most immediate next step, this is just documentation deep, signalling that this is deprecated and seeing whether and what reactions we get. :smile:

PR #836, opened 2018-06-27
WIP: Guest mode removal

Doug-Reed:
guestMode is no longer tracked in $rootScope. Individual controllers call the service as needed. 

----

### DIFF
--- a/web/src/main/webapp/my-app/layout/controllers.js
+++ b/web/src/main/webapp/my-app/layout/controllers.js
@@ -55,6 +55,18 @@ define(['angular', 'jquery'], function(angular, $) {
         };
         $scope.$emit('REMOVE_WIDGET', data);
       };
+
+      $scope.guestMode = function(mainService) {
+        mainService.isGuest()
+          .then(function(isGuest) {
+            if (isGuest) {
+              return true;
+            }
+            return isGuest;
+        }).catch(function() {
+          $log.warn('Cannot get isGuest');
+          return true;
+        });
     }])
 
   /**
@@ -63,9 +75,9 @@ define(['angular', 'jquery'], function(angular, $) {
    * /widget/partials/widget-card.html)
    */
   .controller('WidgetController',
-  ['$controller', '$log', '$scope', '$rootScope', '$mdToast',
+  ['$controller', '$log', '$scope', '$rootScope', '$mdToast', '$mainService',
     '$sessionStorage', '$filter', '$mdColors', 'layoutService',
-    function($controller, $log, $scope, $rootScope, $mdToast,
+    function($controller, $log, $scope, $rootScope, $mdToast, $mainService,
              $sessionStorage, $filter, $mdColors, layoutService) {
       var vm = this;
       $scope.selectedNodeId = '';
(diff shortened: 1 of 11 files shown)


web/src/main/webapp/my-app/layout/spec/layout_controller_spec.js
```diff
@@ -95,10 +95,10 @@ define(['angular-mocks', 'portal', 'my-app'], function() {
 
       it('should set layoutEmpty to true after return empty layout',
         function() {
-          httpBackend.whenGET(groupURL).respond([]);
```
apetro: 8d06bdcbf8af3c94408c36b9bc814969999c1dd5 is a whitespace-only commit. Consider a [commit message more like `style: normalize indent in layoutEmpty true test`](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type), rather than `refactor: test changes`.

Conventional Commits are nothing like sufficient for good software, but they do serve a purpose -- I had to remember to put a `w=1` on the end of the URL to figure out that this commit doesn't include changes to the test.

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -366,6 +370,23 @@ define(['angular', 'jquery'], function(angular, $) {
             $sessionStorage.layout = layoutService.getLayout();
           });
       };
+      vm.setGuestMode = function() {
+        $scope.guestMode = false;
+        layoutService.getGuestMode().then(function(result) {
+          if (angular.isDefined(result) && !result) {
```
apetro: Would this be more readable as an `if...else if` sequence? something like

```
if (result) {
  $scope.guestMode = true;
   return true;
} else if (angular.isDefined(result)) {
  $scope.guestMode = false;
  return false;
} else {
  return result;
}
```

How about reducing the number of places this function returns, would this be an improvement?

```
if (result) {
  $scope.guestMode = true;
} else if (angular.isDefined(result)) {
  $scope.guestMode = false;
} 

return result;
```

Or what about reducing the number of conditions further? Since it defaults `$scope.guestMode = false` presumably it doesn't need to do anything unless `$scope.guestMode` instead needs to be set to true?

```
vm.setGuestMode = function() {
  $scope.guestMode = false;

  layoutService.getGuestMode().then(function(result) {
    if (result) {
      $scope.guestMode = true;
    }
    return result;
  })
  .catch(function() {
    $log.warn('could not retrieve guest mode');
   });
};
```

web/src/main/webapp/my-app/layout/spec/layout_controller_spec.js
```diff
@@ -35,6 +35,7 @@ define(['angular-mocks', 'portal', 'my-app'], function() {
 
       var q;
       var deferred;
```
apetro: Maybe rename to `deferredLayout` for parallelism?

web/src/main/webapp/my-app/layout/controllers.js
```diff
@@ -388,19 +388,18 @@ define(['angular', 'jquery'], function(angular, $) {
         $rootScope.layout == null) {
           $rootScope.layout = [];
           $scope.layoutEmpty = false;
+          // Get user's home layout
+          layoutService.getLayout().then(function(data) {
+            $rootScope.layout = data.layout;
+            $scope.layout = data.layout;
+            if (data.layout && data.layout.length == 0) {
```
ChristianMurphy: :art: `data.layout.length === 0`

PR #843, opened 2018-07-11
fix: rm app directory search results count badge, temporarily

apetro:
Mitigates bug in which the badge indicates more results than are visible to user, by removing the unreliable badge.

no-app-directory-results-badge

Intended as temporary workaround. Suppresses a misleading badge until deeper fix to reliably show the correct number in the badge. 

Hypothesis: underlying bug is it's including in the badge count items on which user has `BROWSE` but not `SUBSCRIBE` whereas the user-visible results listing is only showing items on which user has both kinds of permission.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to
 
 ### Fixed
 
+* Removes app directory search results count badge. Intended as temporary
+  mitigation for bug wherein the badge did not reliably show a correct count.
+  (#843)
 * Successfully launch searches into today.wisc.edu from Madison's theme (#829)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ and this project adheres to
 
 ### Fixed
 
+* Removes app directory search results count badge. Intended as temporary
+  mitigation for bug wherein the badge did not reliably show a correct count.
+  (#843)
 * Successfully launch searches into today.wisc.edu from Madison's theme (#829)
 * May have fixed a bug where app directory search result badge could get out of
```
apetro: ( #842 would remove this prematurely optimistic changelog entry).

PR #844, opened 2018-07-13
Show app directory search results the user can BROWSE but not SUBSCRIBE

apetro:
Restores the search result count badge on the app directory results tab. That badge had been temporarily removed in #843 because it was including items the user could BROWSE-but-not-SUBSCRIBE in the count even though those search results were not displayed to the user.

This makes that badge count correct by following through on including the BROWSE-but-not-SUBSCRIBE items in the search results display. These you're-allowed-to-learn-about-this-item-but-you-cannot-actually-launch-it items are displayed at the end of the search results, after the items the user can actually launch.

include-not-subscribable-in-search-results

(The use case here is there may be situations in which it is helpful for a user to know about, reference, show another user about an item in MyUW, even though that item isn't intended for their own use. For instance, academic advisors helping students to understand content available to students, even though the advisors themselves cannot launch all of that content.)

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ and this project adheres to
 
 ### Deprecated
 
+## [8.3.2][] - 2018-08-01
+
+### Fixed
+
+* Compact widgets no longer disappear when reordering via click-and-drag (#848)
+
 ## [8.3.1][] - 2018-07-26
 
 ### Updated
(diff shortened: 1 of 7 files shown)


davidmsibley:
What are we waiting for on this PR?

apetro:
We are probably waiting on teeing up local data changes so that the `BROWSE` permissions result in users finding content that they might value finding. I should have some local merge requests soon proposing applying the BROWSE permission grants explicitly on the individual `portlet-definition` entity files rather than inheriting these via categories.

PR #864, opened 2018-11-15
MUMMNG-4432: replace sad bucky

Doug-Reed:
Sad Bucky is not our image to use. We have run roughshod over the intellectual property of others. Replacing with a generic. 

----


apetro:
Alright, now that we've been bitten by this: why do we believe we're allowed to use that exclamation point and where do we document that?

Doug-Reed:
image

Public domain.

PR #910, opened 2020-04-09
Update Releasing uPortal-home documentation

nogalpaulina:


docs/releasing.md
```diff
@@ -6,8 +6,68 @@ This project has [adopted Apache rules][] with necessary or pragmatic local adap
...
+
+If you call canonical something else (like, say, `upstream`), you'll need to
+adjust the git push commands accordingly.
+
+Then,
+
++ [Release Maven artifacts][] from
+  [central staging repository][central Maven staging repository]
+
+### Node release
```
apetro, line 48: I don't think we typically do a Node release of uPortal-home. [It'd be `@uportal/home`](https://github.com/uPortal-Project/uportal-home/blob/42321657546e883309bf41a133caa97f47b5bc51/package.json#L2) if we did release it, but [`@uportal/home` was last released on npm 3 years ago](https://www.npmjs.com/package/@uportal/home). The project had the `private` flag set a couple years ago, it looks like [intentionally to prevent publication to npm](https://github.com/uPortal-Project/uportal-home/commit/1fd512dbce00c01765d6097330b8e29a95b11a8f).

nogalpaulina: I do recall getting the privacy error after trying to run `npm publish`, so I didn't include that particular step. But that makes sense - I'll remove the Node release part.

PR #932, opened 2021-09-20
Ensure there are no duplicate fnames in dataLayout

nogalpaulina:
Checking for duplicate elements in data.layout array, filtering duplicate elements out, and saving filtered array as a `filteredDataLayout`.

----

Review for security considerations

- [x] The change has been examined for security impact.

### DIFF
--- a/web/src/main/webapp/my-app/layout/services.js
+++ b/web/src/main/webapp/my-app/layout/services.js
@@ -50,7 +50,6 @@ define(['angular', 'jquery'], function(angular, $) {
               var successFn;
               var errorFn;
               var defer;
-
               // first, check the local storage...
               if (data) {
                 $log.log('getLayout (new version): found layout in cache: ' + data);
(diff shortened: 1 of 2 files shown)


web/src/main/webapp/my-app/layout/services.js
```diff
@@ -380,8 +384,8 @@ define(['angular', 'jquery'], function(angular, $) {
               }
               result.layout = portlets;
             }
-          } else if ($.isArray(data.layout)) { // layoutDoc
-            result.layout = data.layout;
+          } else if ($.isArray(filteredDataLayout)) { // layoutDoc
```
apetro, line 387: The if (line 353) / else if  (365) / else if (387) is supporting 3 different layout APIs. Three different representations of layout. Only the last is a straight up array of fnames. So the filtering the array of fnames to remove duplicates should probably be inside this third branch of the if rather than before the if. That is, `data.layout` is not reliably an array of fnames until inside this else.

Alternatively, we could get confident that data.layout really is always an array of fnames, simplify away the if.

apetro:
Let's try it.

